PR #4, opened 2016-05-13
Replace ./update-index.sh with ./update + Replace list.js with list.json

Latrasis:
In request for: https://github.com/ethereum/solc-js/issues/10#issuecomment-218628533


chriseth:
Thanks for your contribution! If I understand this correctly, it is about

1. generating a version list in json format
2. providing the filtered list pre-generated

While pre-filtering the release versions is a good idea, I see some problems with the specific solution at hand:

It adds nodejs as a build requirement which we currently do not want to add and it breaks backwards-compatibility. I think it would be good to have a separate json-file that contains the versions and the pre-filtered versions, but is generated via a bash script. `list.js` should stay, as should `list.txt` (in its current sorting order).

Latrasis:
Ok i have some solutions in my head, but first some questions:

1. Can I ask what is the usecase for `list.txt`? It's redundant considering the `bin` file can simply be listed.
2. Why is the latest version stored in both the root directory and the `bin` directory?
3. Why store the `list.js` in `bin`?

As for a solution:
Considering that the only thing being updated currently is pushing another version to `bin`. Why delegate the sorting to `solc-bin?` The sorting can be done by `solc-js` by quering the `bin` folder directory remotely and parse the `bin/*.js` dynamically with a filtered list.

So considering the steps on a new update:
1. Jenkins pushes a new version to `bin/`.
2. Solc queries the `...bin/`directory with http, parses the version list, and picks the latest version.

axic:
Perhaps `list.json` could have two sections:
- releases
- builds (including releases) - matches the current list.js, in the same order

chriseth:
You cannot simply list the directory contents, at least not via ajax requests to github (e.g. from browser-solidity). That's why there is `list.txt`. "latest" is stored in bin and in the root directory for backwards-compatibility reasons. I guess there are reasons for and against storing the list in the bin directory, but we should not change it, also for backwards-compatibility reasons.

Since we already have those lists, they can simply be used by solc-js. I'm fine with adjusting the bash script so that sorting and filtering is done by it.

axic:
@Latrasis would you be able to change the bash script to do the sorting and generate the JSON file? (If @chriseth is absolutely positive nodejs as a requirement is too much for the build system).

chriseth:
@Latrasis @axic actually I am just in the process of migrating the creation of binaries to a system where we can have nodejs - so no need to convert it to a bash script.

axic:
@chriseth Is there any progress in this direction? Having a `list.json` available would make the offline-mode of browser-solidity much more cleaner than it is currently.

chriseth:
Yes, we can now use nodejs to generate the files.

Changes still required here:

1) Please do not change the sorting order of the txt file.
2) (no change) .js file can be replaced by .json file
3) If we change the format anyway, it would be good to somehow get the list of all "released" versions from the json file.

axic:
@chriseth great news!

@Latrasis can you please rebase and update your changes to reflect these restrictions?

Latrasis:
@axic: Sure thing, I should get it done sometime tomorrow.

axic:
@Latrasis thank you!

axic:
Also please run `semistandard` on the file to follow the general indent rules the projects are using.

Latrasis:
Sorry for the delay! Here's the fix.

axic:
Thank you!

Can you please do a rebase to not have the merge commit?

Also, I think it would make sense changing `list.json`:
- rename `allVersions` to `builds`
- add `build` field next to `version` to include the full build information (ie. `v0.3.2-2016-05-06-9e36bdd`)

And for backwards compatibility probably we should add back `list.js`.
```

Latrasis:
Since the this branch is kinda a mess, i'll do a fresh fork instead.

PR #5, opened 2016-08-15
Generate list.json as well

axic:
This is cherry-picked from @Latrasis' PR: https://github.com/ethereum/solc-bin/pull/4

Still need to add this:

> Also, I think it would make sense changing list.json:
> - rename allVersions to builds
> - add build field next to version to include the full build information (ie. v0.3.2-2016-05-06-9e36bdd)


update
```diff
@@ -0,0 +1,95 @@
...
+(function() {
+  var version = '';
+  for (var i = soljsonSources.length - 1; i >= 0; --i) {
+    var thisVersion = soljsonSources[i].match(/soljson-v([0-9.]*)-.*.js/)[1];
+    if (thisVersion === version)
+      continue;
+    version = thisVersion;
+    soljsonReleases[version] = soljsonSources[i];
+  }
+})();
```
axic: Should actually generate the `soljsonReleases` map instead of including this code.

update
```diff
@@ -0,0 +1,95 @@
...
+    .filter(function (version) { return version })
+    .map(function (pars) { return { path: pars[0], version: pars[1] } })
+
+  const parsedObject = parsedList
+    .reduce(function (prev, next) {
+      prev[next.version] = next.path
+      return prev
+    }, {})
+
+  const pathList = parsedList
```
axic: Rename these to more meaningful variables:
- `builds` (all builds)
- `releases`

chriseth:
What is the status of this one? In the future, the version strings will be proper semver versions with pre-release syntax. The develop branch (nightly builds) will always have a bumped minor version with prerelease specifier, only the release branch will have a proper non-pre-release version. This script should be able to cope with that and assign releaseVersions properly.

axic:
@chriseth I've actually made some updates this morning. Will push it soon.

axic:
@chriseth uploaded, please review. It seems a bit messy, because everything is written in the same order as it was before. The `run update` commit is the output of the script.

update
```diff
@@ -0,0 +1,82 @@
...
+const path = require('path')
+
+// This script updates the index files list.js and list.txt in the bin directory,
+// as well as the soljson-latest.js files.
+
+fs.readdir(path.join(__dirname, '/bin'), function (err, files) {
+  if (err) return err
+
+  const parsedList = files
+    .map(function (file) { return file.match(/^soljson-v([0-9.]*)-(.*).js$/) })
```
chriseth: This will be different with future versions, it will actually be a proper semver version string, and the semver comparator can be used as is.

chriseth:
@axic we can also just rename all the files and do proper semver comparison and filtering.

PR #6, opened 2016-08-31
list.json to include builds next to releases

axic:
This was forgotten in #5.

The format is:
```
{
  "builds": [
    {
      "path": "soljson-v0.1.1-2015-08-04-6ff4cd6.js",
      "version": "0.1.1",
      "build": "2015-08-04-6ff4cd6"
    },
    {
      "path": "soljson-v0.1.2-2015-09-14-d0d36e3.js",
      "version": "0.1.2",
      "build": "2015-09-14-d0d36e3"
    },
    ...
  ],
  "releases": {
    "0.3.6": "soljson-v0.3.6-2016-08-10-3fc68da.js",
    "0.3.5": "soljson-v0.3.5-2016-06-10-5f97274.js",
    "0.3.4": "soljson-v0.3.4-2016-05-31-7dab890.js",
   ...
   ]
}
```


axic:
@chriseth I think we should merge this ASAP

update
```diff
@@ -40,7 +40,7 @@ fs.readdir(path.join(__dirname, '/bin'), function (err, files) {
   })
 
   // Write bin/list.json
-  fs.writeFile(path.join(__dirname, '/bin/list.json'), JSON.stringify(releases), function (err) {
+  fs.writeFile(path.join(__dirname, '/bin/list.json'), JSON.stringify({ builds: parsedList, releases: releases }, null, 2), function (err) {
```
chriseth: This comment is about line 17: shouldn't it be "prerelease" instead of "build"?

axic: `build` contains every single build (including releases), with complete info (version and build tag). Should it have a tag for `release: true`?

chriseth: No, I mean what currently is the build tag should be the prerelease tag.

axic: It contains both prerelease and release versions atm.

axic:
@chriseth:

Would this suit the new versioning system?
```
{
  "builds": [
    {
      "path": "soljson-0.3.6-develop.2016.08.31+commit.2240adcf-mod.Linux.clang.js",
      "version": "0.3.6",
      "major": 0,
      "minor": 3,
      "patch": 6,
      "prerelease": "develop.2016.08.31",
      "build": "commit.2240adcf-mod.Linux.clang"
    },
    {
      "path": "soljson-0.3.6+commit.8842001-mod.Linux.clang.js",
      "version": "0.3.6",
      "major": 0,
      "minor": 3,
      "patch": 6,
      "build": "commit.8842001-mod.Linux.clang"
    },
    ...
  ],
  "releases": {
    "0.3.6": "soljson-0.3.6+commit.8842001-mod.Linux.clang.js",
   ...
   ]
}
```

chriseth:
What is the purpose of the various fields? If you have a semver parser, most of the information is redundant. In other words: When would you be interested in the detailed version information and do not have a semver parser?

axic:
If we're preparing a JSON file with releases, and a list of all builds, then we can do the semver parsing at the same time (because we'll do it anyway).

The consumer of this JSON doesn't need to do that.

chriseth:
The consumer of the json has to parse the json, so some parsing is required. All I'm saying is that this json file will be much larger than necessary.

axic:
To be clear, with the change to semver, it indeed does seem a bit duplicate. Before changing to semver, it is pretty much needed.

I think the version and the prerelease fields are useful though. The rest can be dropped.

axic:
Even with semver, I think the filename and the version should be two different fields, so that a consumer doesn't need to understand the filename schema.

chriseth:
ok!

chriseth:
Oh, another thing: Immediately after 0.4.0 is released, the develop version will jump to `0.5.0-nightly....`. Then the next release might be `0.4.1`, but develop will stay at `0.5.0-nightly...`. Does that generate any problems here?

axic:
I think a better approach is to assume only a patch release is happening, i.e. after 0.4.0, the nightly is 0.4.1. If there's a change merged, which would cause a minor bump, the nightly jumps to 0.5.0. Again, if it warrants a major change, it jumps to 1.0.0.

chriseth:
I think it is safer to jump to `0.5.0-nightly`, but anyway, this system should be able to comply with that.

axic:
@chriseth renamed `build` to `prerelease`

PR #8, opened 2016-10-10
Sort using semver

axic:
Fixes the out of order versions.

However for simplicity we should just move to dates in the form of YYYY.MM.DD (and not YYYY.M.D).


update
```diff
@@ -4,17 +4,33 @@
...
 fs.readdir(path.join(__dirname, '/bin'), function (err, files) {
   if (err) return err
 
+  function buildVersion (build) {
+    var version = build.version;
+    if (build.prerelease && build.prerelease.length > 0) {
+      version += '-' + build.prerelease;
+    }
+    if (build.build && build.build.length > 0) {
+      version += '-' + build.build;
```
chriseth: Should be `'+'` here.

update
```diff
@@ -4,17 +4,33 @@
...
 const path = require('path')
+const semver = require('semver')
 
 // This script updates the index files list.js and list.txt in the bin directory,
 // as well as the soljson-latest.js files.
 
 fs.readdir(path.join(__dirname, '/bin'), function (err, files) {
   if (err) return err
 
+  function buildVersion (build) {
```
chriseth: Do we really need to rebuild it again?

update
```diff
@@ -4,17 +4,33 @@
...
+  }
+
   const parsedList = files
     .map(function (file) { return file.match(/^soljson-v([0-9.]*)(-([^+]*))?(\+(.*))?.js$/) })
     .filter(function (version) { return version })
-    .map(function (pars) { return { path: pars[0], version: pars[1], prerelease: pars[3], build: pars[5] } })
+    .map(function (pars) { console.log(pars[0]); return { path: pars[0], version: pars[1], prerelease: pars[3], build: pars[5] } })
+    .sort(function (a, b) {
+      // NOTE: a vs. b (the order is important), because we want oldest first in the list
+      return semver.compare(buildVersion(a), buildVersion(b))
```
chriseth: We could just use the semver module to also parse the version string.

chriseth:
Haha! Yes, we could use `yyyy-mm-dd`, but that would not be semver, because leading zeros are not allowed, not even in the prerelease components....

update
axic: Actually if we rebuild it with MM.DD we don't even need semver to do the comparison.

update
axic: Actually semver doesn't seem to return the `build` part (only versions and `prerelease`).

axic:
This works well with #11 (parsedList becomes the sortedList).

update
axic: Fixed.

axic:
@chriseth are we happy adding `semver` a requirement of update?

The Solidity push script
a) needs to be changed to include `npm install semver` or
b) we need to add package.json and the script needs to be change to include `npm install`

I prefer a) for the moment, until we add more dependencies.

axic:
Actually I think it is more future proof (less changes in the Solidity repo) if we have a package.json (marked `private`).

@chriseth added that in that case we could also run `npm run build` in Solidity and therefore making the Solidity repo even less dependent on the layout of this repo.

PR #10, opened 2016-10-11
Write latestRelease field to list.json

axic:
This makes it much easier to consume list.json: `list.releases[list.latestRelease]`


axic:
Need to merge #11 first and this needs to be adjusted after.

PR #15, opened 2016-12-01
Include keccak256 for every compiler

axic:

PR #16, opened 2016-12-01
Include Swarm hash for every compiler

axic:
This is based on #15.


axic:
I think we should merge this two changes (#15 and this), because that is one step to do "authenticated binaries".

axic:
I'll update travis to include a C compiler because its needed for the SHA3 dependency in node.js

PR #17, opened 2017-01-02
Revert "Include Swarm hash for every compiler"

chriseth:
Reverts ethereum/solc-bin#16

This causes the deployment to fail, because `Buffer.alloc` is not available in the node version used on travis.

Reverting until this is resolved.


axic:
How come this became an issue? It was working when merged.

axic:
@chriseth just include this in `solidity/.travis.yml` and roll this back please:

```
node_js:
- stable
```

It should have been fixed at the source :)

PR #19, opened 2018-03-28
Include github URLs in list.json

axic:
After this change it is possible to rely only on `list.json` in a client. The client can look for `http://` or `https://` depending on what they run on.

### DIFF
--- a/bin/list.json
+++ b/bin/list.json
@@ -5,9 +5,12 @@
       "version": "0.1.1",
       "build": "commit.6ff4cd6",
       "longVersion": "0.1.1+commit.6ff4cd6",
+      "sha256": "0xfc54135edfe07d835216f87fa6c3b9049e1415c2825027b926daad018e09269f",
       "keccak256": "0xd8b8c64f4e9de41e6604e6ac30274eff5b80f831f8534f0ad85ec0aff466bb25",
       "urls": [
-        "bzzr://8f3c028825a1b72645f46920b67dca9432a87fc37a8940a2b2ce1dd6ddc2e29b"
+        "bzzr://8f3c028825a1b72645f46920b67dca9432a87fc37a8940a2b2ce1dd6ddc2e29b",
+        "https://ethereum.github.io/solc-bin/bin/soljson-v0.1.1+commit.6ff4cd6.js",
+        "http://ethereum.github.io/solc-bin/binsoljson-v0.1.1+commit.6ff4cd6.js"
       ]
     },
     {
@@ -15,9 +18,12 @@
       "version": "0.1.2",
       "build": "commit.d0d36e3",
       "longVersion": "0.1.2+commit.d0d36e3",
+      "sha256": "0x7ad9fa9de246a33c5e5472127b6e0b6e713f3900c7ea360c7c2824f6e9202a0f",
       "keccak256": "0xa70b3d4acf77a303efa93c3ddcadd55b8762c7be109fd8f259ec7d6be654f03e",
       "urls": [
-        "bzzr://e662d71e9b8e1b0311c129b962e678e5dd63487ad9b020ee539d7f74cd7392c9"
+        "bzzr://e662d71e9b8e1b0311c129b962e678e5dd63487ad9b020ee539d7f74cd7392c9",
+        "https://ethereum.github.io/solc-bin/bin/soljson-v0.1.2+commit.d0d36e3.js",
+        "http://ethereum.github.io/solc-bin/binsoljson-v0.1.2+commit.d0d36e3.js"
       ]
     },
     {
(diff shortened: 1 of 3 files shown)


update
```diff
@@ -39,8 +39,13 @@ fs.readdir(path.join(__dirname, '/bin'), function (err, files) {
     .map(function (pars) {
       const fileContent = readFile(pars.path)
       pars.longVersion = buildVersion(pars)
+      pars.sha256 = '0x' + ethUtil.sha256(fileContent).toString('hex')
       pars.keccak256 = '0x' + ethUtil.sha3(fileContent).toString('hex')
-      pars.urls = [ 'bzzr://' + swarmhash(fileContent).toString('hex') ]
+      pars.urls = [
+        'bzzr://' + swarmhash(fileContent).toString('hex'),
+        'https://ethereum.github.io/solc-bin/bin/' + pars.path,
+        'http://ethereum.github.io/solc-bin/bin' + pars.path
```
axic: Actually since we have `solc-bin.ethereum.org` as a redirect there, should we use that?

axic:
ping @chriseth

update
chriseth: Why do we need absolute links here? This file is just a summary of the other files in the same location.

axic: It would work if we consider the current approach with github pages, however I was thinking about having the "manifest" (`list.json`) available on swarm or ipfs and as such it would need absolute URLs.

We could include a relative path as an option in the URLs though.

chriseth: Actually if you use a swarm manifest, that would be an argument to use relative paths (which are then valid both inside swarm and inside github pages).

axic: Oh btw it is a URL and not a path. Not sure how to represent relative paths on URLs.

chriseth: We already have the `path` member, so actually no need to put them here at all.
We could add ipfs urls, though...

ericramos1980 (approved):

cameel:
@axic Maybe, instead of adding new URL variants to the list in every dict we should rather add a separate, top-level key with official sources? The number of sources will probably keep getting bigger. We already want to add IPFS so that'll be 4 items per file. The dicts already contain enough information to construct URLs if you know the source.

Also, I think that having a list of official sources (either in the file, or maybe just in README) would have helped us with the transition to S3. `solc-bin.ethereum.org` will keep working after we switch to S3 for hosting but we want to disable GH pages (i.e. `ethereum.github.io`). This will break tools that depend on the latter. They shouldn't depend on it but I don't know if it has been communicated clearly enough so they might. If there was a list, it would have been more obvious.

chriseth:
I would prefer it if this is really an index of the current list - and thus does not depend on or contain any information about where it is located.

cameel:
How about including a separate, static file in the root of the repository, with a list of possible source URLs?

update
axic: I think we actually want to add the official `https://binaries.soliditylang.org/...` URL.

Reason: all the URLs we have in this array are pointing to a specific file and are not paths within a manifest/collection. If we publish `list.json` + the paths as a collection as well, for that use case the consumer can try the path, otherwise if `list.json` is just independently published on IPFS or swarm they can still opt to acquire the files via various means.

bin/list.json
```diff
@@ -5,19 +5,25 @@
       "version": "0.1.1",
       "build": "commit.6ff4cd6",
       "longVersion": "0.1.1+commit.6ff4cd6",
+      "sha256": "0xfc54135edfe07d835216f87fa6c3b9049e1415c2825027b926daad018e09269f",
       "keccak256": "0xd8b8c64f4e9de41e6604e6ac30274eff5b80f831f8534f0ad85ec0aff466bb25",
       "urls": [
-        "bzzr://8f3c028825a1b72645f46920b67dca9432a87fc37a8940a2b2ce1dd6ddc2e29b"
+        "bzzr://8f3c028825a1b72645f46920b67dca9432a87fc37a8940a2b2ce1dd6ddc2e29b",
+        "https://ethereum.github.io/solc-bin/bin/soljson-v0.1.1+commit.6ff4cd6.js",
+        "http://ethereum.github.io/solc-bin/binsoljson-v0.1.1+commit.6ff4cd6.js"
```
chriseth, line 13: Shouldn't this be
```suggestion
        "http://ethereum.github.io/solc-bin/bin/soljson-v0.1.1+commit.6ff4cd6.js"
```
?

Mhmonicox (approved):

Mahony36 (approved):

Mahony36 (approved):

Varshans2 (approved):
good

r0qs:
@axic @cameel should we close this? And instead work in the other suggested approaches here: https://github.com/ethereum/solc-bin/pull/19#issuecomment-654357849 and here: https://github.com/ethereum/solidity/issues/2238#issuecomment-832314927

bin/list.json
mehdijalaly: 0xa1291d7e34614eb11AD305f299f062cAB485e0E5

MaaZzk (approved):

curtisdow1973-sys (approved):

curtisdow1973-sys (approved):

curtisdow1973-sys (approved):

Varshans2 (approved):

Varshans2 (approved):

PR #20, opened 2018-04-02
Sort pre-0.4.0 versions properly

axic:
This fixes the sorting order for <0.4.0 -- which was changed by #8.


axic:
@cameel while this is definitely not a pressing issue, perhaps we may want to get this done?

update
```diff
@@ -21,6 +21,13 @@ fs.readdir(path.join(__dirname, '/bin'), function (err, files) {
     if (build.prerelease && build.prerelease.length > 0) {
       version += '-' + build.prerelease
     }
+
+    // workaround for pre-0.4.0, where prereleases didn't bump the version
+    // NOTE: the text must be lexically after 'nightly'
+    if (semver.major(build.version) === 0 && semver.minor(build.version) < 4) {
```
cameel:
```suggestion
    if (semver.lt(build.version, '0.4.0')) {
```

update
```diff
@@ -21,6 +21,13 @@ fs.readdir(path.join(__dirname, '/bin'), function (err, files) {
...
+      version += '-workaround'
```
cameel: I think it would be better to change the order in the file without actually modifying the content of the fields in JSON.

cameel (commented):
I'm not sure there's any point in doing this given that there are no nightlies earlier than 0.4.10 in the repo. Or do we want to restore them?

update
axic: Which file?

cameel: Ah, sorry, I misread it. I thought it changes the `version` field value that goes into the JSON but now I see that this is just a local variable that is returned from the function.

axic: Yes, it is a trick for the internal `sort()` to work in a way we want.

1alireza32109:
> This fixes the sorting order for <0.4.0 -- which was changed by #8.

axic:
> I'm not sure there's any point in doing this given that there are no nightlies earlier than 0.4.10 in the repo.

They definitely existed, hence the PR.

> Or do we want to restore them?

Hmm, I think we should have the proper history, especially as we now serve this off-github.

cameel:
> Hmm, I think we should have the proper history, especially as we now serve this off-github.

OK then. Here's a PR that restores them: https://github.com/ethereum/solc-bin/pull/103.

mrx23dot:
url https://solc-bin.ethereum.org/windows-amd64/list.json only goes back to v0.4.1

where  https://api.github.com/repos/ethereum/solidity/releases?per_page=1000
has v0.1.2

**There are still live contracts on the network that use old versions like v0.2.2-2016**
and security experts still want to verify/review old contracts for which they need old compilers.

cameel:
@mrx23dot We have emscripten binaries going back as far as 0.1.1: https://github.com/ethereum/solc-bin/tree/gh-pages/emscripten-asmjs. These work on any platform.

By the way this is unrelated to the PR. If you want to request something related to solc-bin binaries, please [open an issue in the solidity repo](https://github.com/ethereum/solidity/issues/new).

r0qs (approved):

cameel:
@r0qs The PR is incomplete - the lint job is failing. Since #103 was merged, the old nightlies have been restored but lists need to be regenerated with the new sort order.

emscripten-wasm32/list.json
```diff
@@ -4,7 +4,20 @@
...
+        "bzzr://83e99aa35ae67e71bf77040e5b4686aeb3558919f172b6c28213a4dcd8d79e91",
+        "dweb:/ipfs/Qme9brfZS3XhbiRbbNDKhBpgFknyD92omMmYa7XSf56bJP"
+      ]
+    },
+    {
+      "path": "solc-emscripten-wasm32-v0.3.6-workaround+commit.3fc68da5.js",
+      "version": "0.3.6",
+      "prerelease": "workaround",
+      "build": "commit.3fc68da5",
+      "longVersion": "0.3.6-workaround-workaround+commit.3fc68da5",
```
r0qs, line 20: @cameel although the tests passed, I noticed one weird case here though. So please do not approve it yet. It seems that the symlink is being created with the `-workaround` and then the `longVersion` get it doubled.

PR #21, opened 2018-06-27
Insert reportedVersion in list.json

axic:
Part of https://github.com/ethereum/solc-js/issues/231

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   },
   "devDependencies": {
     "ethereumjs-util": "^5.0.1",
+    "solc": "^0.4.24",
     "standard": "^8.4.0",
     "swarmhash": "^0.1.0"
   },
--- a/package.json
+++ b/package.json
@@ -8,7 +8,6 @@
   },
   "devDependencies": {
     "ethereumjs-util": "^5.0.1",
-    "solc": "^0.4.24",
     "standard": "^8.4.0",
     "swarmhash": "^0.1.0"
   },
--- a/update
+++ b/update
@@ -7,7 +7,7 @@ const path = require('path')
 const semver = require('semver')
 const ethUtil = require('ethereumjs-util')
 const swarmhash = require('swarmhash')
-const solc = require('solc/wrapper')
+const cp = require('child_process')
 
 // This script updates the index files list.js and list.txt in the bin directory,
 // as well as the soljson-latest.js files.
@@ -35,7 +35,10 @@ dirs.forEach(function (dir) {
     }
 
     function readBuiltinVersion (file) {
-      return solc(require(path.join(__dirname, '/bin', file))).version()
+      // NOTE: should be using this, but it leaks memory
+      // return solc(require(path.join(__dirname, '/bin', file))).version()
+      const filename = path.join(__dirname, '/bin', file)
+      return cp.execSync(`/usr/bin/env node -e "console.log(require('${filename}').cwrap('version', 'string', [])())"`).toString().trim()
     }
 
     // ascending list (oldest version first)
--- a/update
+++ b/update
@@ -7,6 +7,7 @@ const path = require('path')
 const semver = require('semver')
 const ethUtil = require('ethereumjs-util')
 const swarmhash = require('swarmhash')
+const solc = require('solc/wrapper')
 
 // This script updates the index files list.js and list.txt in the bin directory,
 // as well as the soljson-latest.js files.
@@ -33,6 +34,10 @@ dirs.forEach(function (dir) {
       return fs.readFileSync(path.join(__dirname, dir, file))
     }
 
+    function readBuiltinVersion (file) {
+      return solc(require(path.join(__dirname, '/bin', file))).version()
+    }
+
     // ascending list (oldest version first)
     const parsedList = files
       .map(function (file) { return file.match(/^soljson-v([0-9.]*)(-([^+]*))?(\+(.*))?.js$/) })
@@ -41,6 +46,7 @@ dirs.forEach(function (dir) {
       .map(function (pars) {
         console.log('Processing ' + pars.path)
         const fileContent = readFile(pars.path)
+        pars.reportedVersion = readBuiltinVersion(pars.path)
         pars.longVersion = buildVersion(pars)
         pars.keccak256 = '0x' + ethUtil.sha3(fileContent).toString('hex')
         pars.urls = [ 'bzzr://' + swarmhash(fileContent).toString('hex') ]
--- a/update
+++ b/update
@@ -39,6 +39,7 @@ dirs.forEach(function (dir) {
       .filter(function (version) { return version })
       .map(function (pars) { return { path: pars[0], version: pars[1], prerelease: pars[3], build: pars[5] } })
       .map(function (pars) {
+        console.log('Processing ' + pars.path)
         const fileContent = readFile(pars.path)
         pars.longVersion = buildVersion(pars)
         pars.keccak256 = '0x' + ethUtil.sha3(fileContent).toString('hex')
--- a/update
+++ b/update
@@ -36,8 +36,8 @@ dirs.forEach(function (dir) {
 
     function readBuiltinVersion (file) {
       // NOTE: should be using this, but it leaks memory
-      // return solc(require(path.join(__dirname, '/bin', file))).version()
-      const filename = path.join(__dirname, '/bin', file)
+      // return solc(require(path.join(__dirname, dir, file))).version()
+      const filename = path.join(__dirname, dir, file)
       return cp.execSync(`/usr/bin/env node -e "console.log(require('${filename}').cwrap('version', 'string', [])())"`).toString().trim()
     }
 
--- a/update
+++ b/update
@@ -38,7 +38,7 @@ dirs.forEach(function (dir) {
       // NOTE: should be using this, but it leaks memory
       // return solc(require(path.join(__dirname, dir, file))).version()
       const filename = path.join(__dirname, dir, file)
-      return cp.execSync(`/usr/bin/env node -e "console.log(require('${filename}').cwrap('version', 'string', [])())"`).toString().trim()
+      return cp.execSync(`/usr/bin/env node -e "var solc = require('${filename}'); console.log(solc.cwrap(('_solidity_version' in solc) ? 'solidity_version' : 'version', 'string', [])())"`).toString().trim()
     }
 
     // ascending list (oldest version first)


fanatid:
@axic try with `execSync`, example:
```js
const cp = require('child_process')
console.log(cp.execSync(`node -e "console.log(require('./bin/soljson-v0.3.1-nightly.2016.4.7+commit.54bc2a').cwrap('version', 'string', [])())"`).toString())
```

axic:
Thanks! That was my idea for workaround too, but would be nice to figure out what is leaking memory.

axic:
@fanatid @chriseth can you review?

package.json
```diff
@@ -7,6 +7,7 @@
     "semver": "^5.3.0"
   },
   "devDependencies": {
+    "child_process": "^1.0.2",
```
fanatid: Why? `child_process` is of the node core modules: https://nodejs.org/api/child_process.html

fanatid:
29min in Travis -- it's crazy. Also, do you need to run `npm run update` locally every time on adding new binary?

axic:
> Also, do you need to run npm run update locally every time on adding new binary?

Yeah, but it is only done on releases and for every nightly.

fanatid:
If Travis should only verify that files is OK, we can speed-up it with matrix build. But if you are ok with 30min on CI, changes is not worth it.
Regarding running it locally, if you need I'd happy push PR with code for multi-process update. Let me know.

chriseth (approved):

axic:
The list must be updated before merging.

update
```diff
@@ -33,7 +33,10 @@ fs.readdir(path.join(__dirname, '/bin'), function (err, files) {
   }
 
   function readBuiltinVersion (file) {
-    return solc(require(path.join(__dirname, '/bin', file))).version()
+    // NOTE: should be using this, but it leaks memory
+    // return solc(require(path.join(__dirname, '/bin', file))).version()
+    const filename = path.join(__dirname, '/bin', file)
+    return cp.execSync(`/usr/bin/env node -e "console.log(require('${filename}').cwrap('version', 'string', [])())"`).toString().trim()
```
axic, line 39: An added side-effect of this is it would fail if the binary can't be loaded or is missing a the version function (e.g. incompatible).

chriseth:
Needs rebase.

axic:
@cameel what do you think about this? Since you have been rebuilding binaries. The goal of this was to include the version hardcoded in the binary into the JSON list. Currently (or at the time of this PR) the JSON only contained the normalised version, but certain old compilers do not follow that. Here's a list of discrepancies: https://github.com/ethereum/solc-js/blob/master/translate.js#L3

cameel:
I'm not sure if it's actually useful for users but it would definitely be helpful for us to see these versions somewhere just as a sanity check. Putting them on the list does make them and any changes pretty visible.

At times I did need that info myself and I had to gather it manually. I think that seeing them would also make us spot some problems much earlier: https://github.com/ethereum/solc-bin/pull/64, https://github.com/ethereum/solidity/issues/7512, https://github.com/ethereum/solidity/issues/10183.

update
cameel: This would be a problem because we have one executable (` soljson-v0.4.1-nightly.2016.9.9+commit.79867f49.js`) that crashes when you run it.

cameel:
> Here's a list of discrepancies: https://github.com/ethereum/solc-js/blob/master/translate.js#L3

The list is bigger :) Versions listed there look like they're just examples of each format. Actually all emscripten binaries below 0.4.0 use non-standard version format. And 0.4.0 and 0.4.2 also were built with uncommitted changes so their version strings contain `.mod`. I checked them all at some point - here's the full list of all version strings for emscripten releases and nightlies up to 0.7.3: https://github.com/ethereum/solc-bin/pull/64#issuecomment-711221554

update
```diff
@@ -7,6 +7,7 @@ const path = require('path')
 const semver = require('semver')
 const ethUtil = require('ethereumjs-util')
 const swarmhash = require('swarmhash')
+const cp = require('child_process')
```
cameel, line 10: This is not used that often so I'd use full name for readability:
```suggestion
const child_process = require('child_process')
```
`cp` reads like the command-line copying utility to me :)

axic:
True, that list was the cases which need parsing. And ones with `.mod` still parse with semver, so they never showed up as issues. However you mentioning `.mod` makes me believe this would be actually a very useful feature for sanity check.

update
```diff
@@ -33,13 +34,22 @@ dirs.forEach(function (dir) {
       return fs.readFileSync(path.join(__dirname, dir, file))
     }
 
+    function readBuiltinVersion (file) {
+      // NOTE: should be using this, but it leaks memory
+      // return solc(require(path.join(__dirname, dir, file))).version()
+      const filename = path.join(__dirname, dir, file)
+      return cp.execSync(`/usr/bin/env node -e "var solc = require('${filename}'); console.log(solc.cwrap(('_solidity_version' in solc) ? 'solidity_version' : 'version', 'string', [])())"`).toString().trim()
```
cameel, line 41: It looks like it's going to be very slow. When I was rewriting bytecode comparison scripts and I tried doing something like this, rerunning `node` for each compilation was very slow (https://github.com/ethereum/solidity/issues/10165#issuecomment-744521194). A loop in JS was one or two orders of magnitude faster.

Is it actually a leak or is it just that `update` itself is pretty bad about controlling its memory usage (https://github.com/ethereum/solidity/issues/9956)? Because for the bytecode comparison I run [prepare_report.js](https://github.com/ethereum/solidity/blob/v0.8.2/scripts/bytecodecompare/prepare_report.js) which compiles around 5k test cases in a loop, twice and it does not run out of memory in the [t-bytecode-compare.yml](https://github.com/ethereum/solc-bin/blob/gh-pages/.github/workflows/t-bytecode-compare.yml) action.

update
```diff
@@ -33,13 +34,22 @@ dirs.forEach(function (dir) {
...
+
     // ascending list (oldest version first)
     const parsedList = files
       .map(function (file) { return file.match(/^soljson-v([0-9.]*)(-([^+]*))?(\+(.*))?.js$/) })
       .filter(function (version) { return version })
       .map(function (pars) { return { path: pars[0], version: pars[1], prerelease: pars[3], build: pars[5] } })
       .map(function (pars) {
+        console.log('Processing ' + pars.path)
         const fileContent = readFile(pars.path)
+        pars.reportedVersion = readBuiltinVersion(pars.path)
```
cameel, line 52: One problem with adding it now is that the repo contains binaries for multiple platforms and you won't be able to get versions for all of them in one run. We'd need separate runs for different platforms and then we would have to combine gathered info.

update
axic: Nodejs is bad at controlling memory. We set the loaded emscripten modules to empty reference, but it still won't be GC'd in time somehow. Perhaps it is not a problem anymore with the wasm binaries?

cameel:
Maybe we should have an extra check that newly added binaries do not contain `.mod` in their version string? We could easily do that in `t-bytecode-compare.yml` which already has logic to detect newly added binaries in new PRs (except for Windows ones). We would just have to limit it to versions above 0.4.2.

axic:
> Maybe we should have an extra check that newly added binaries do not contain .mod in their version string? 

I think that would make sense in any case.

update
cameel: ok, right. It's about loading the compiler into memory, not running it. Bytecode comparison loads only one version.

> Perhaps it is not a problem anymore with the wasm binaries?

We still need to run it on all those old asm.js nightlies so I think this does not help us even if it's fixed for wasm. Unless we just compute versions for asm.js once and cache them.

The script now has an option that makes it not recalculate hashes. It could have a similar one for this. The downside would be that if you replace the binary, CI won't update the versions so it would be safer to somehow combine it with detecting which binaries changed in the PR (which adds a bit of complexity to the whole setup).

update
cameel, line 52: I think we need an option for skipping this when running locally, just like you can skip hash recalculation.

update
axic: Lol what? Well, there can be an exception.

cameel: Well, we could, but I'm still not sure if running the executables while generating the list is such a good idea. It's slow enough already and we still have this weird problem with nodejs not exiting immediately.

BTW, here's the error from that executable:
```
TypeError: Invalid Version: 0.4.1-nightly.2016.09.09+commit.79867f49.Emscripten.clang
    at new SemVer (/home/cameel/data/working-set/projects/sol/solc-js/node_modules/semver/semver.js:323:11)
    at compare (/home/cameel/data/working-set/projects/sol/solc-js/node_modules/semver/semver.js:614:10)
    at Function.gt (/home/cameel/data/working-set/projects/sol/solc-js/node_modules/semver/semver.js:643:10)
    at setupMethods (/home/cameel/data/working-set/projects/sol/solc-js/wrapper.js:20:27)
    at Object.<anonymous> (/home/cameel/data/working-set/projects/sol/solc-js/index.js:3:18)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
```
Looking at it now, I see that it's actually a problem in solc-js, not the executable so it should be fixable. The string it reports is probably just not valid semver. It seems to be the `09` in version. Later executables have no zeros and don't crash.

cameel: https://github.com/ethereum/solc-js/issues/562

JRICH00 (approved):

r0qs:
@axic Is this still a thing? I didn't understand the motivation for having the `reportVersion` in `list.json`.

If still relevant, I can finish that PR and also add the flag to ignore it when running locally, as you suggested here: https://github.com/ethereum/solc-bin/pull/21#discussion_r593467151.

However, I don't think we can improve its performance in the way that it is (it took almost one and half hours on my machine, but the out-of-memory error does not happen anymore).
Maybe it would be better to improve the performance of the whole process, i.e., taking over PR https://github.com/ethereum/solc-bin/pull/115 and only after that revisiting this one if it still relevant?

Kurdlove (commented):
best way for this.

update
Kurdlove: very simple.😉

curtisdow1973-sys (approved):

PR #22, opened 2020-01-17
Add webassembly rebuilds.

ekpyron:
For reviewing the changes to the update script I suggest not displaying whiteplace changes.

This depends on us changing the release scripts in the solidity repo to keep generate symlinks for new releases from ``bin`` to ``wasm`` (or the other way around).


ekpyron:
There also needs to be a decision about the leading zero for 0.4.13 and 0.5.14.
The ``wasm`` directory here has them, but so far ``bin`` doesn't.

ekpyron:
I'm changing the order of same-day nightlies in ``bin/list*`` as well... the tests seem to be broken for ``gh-pages`` itself in that regard apparently...

ekpyron:
Honestly I don't know how semver does its sorting for the commit hashes...

ekpyron:
Ah, wait, I need to add 0.6.2 to the ``wasm`` folder as well now. Updating...

ekpyron:
Updated again now including 0.6.2 in the wasm folder.

ekpyron:
Should I change the upload script in the solidity repo to add links to the wasm directory for releases?

chriseth:
Yes, please!

ekpyron:
See https://github.com/ethereum/solidity/pull/8227

PR #23, opened 2020-06-08
Where possible use 8 digit commit hashes and create symlinks for backwards compatibility.

ekpyron:
The only versions for which no proper commit hash is reconstruct here are ``v0.1.1`` and ``v0.1.2``... 

Actually for ``v0.1.1`` the correct hash as per ``--version`` is ``6ff4cd6b``, but it's not part of our repo...
But ``v0.1.2`` reports the weird string ``0.1.2-5c3bfd4b*/.-/clang/int``, so that might mean ``5c3bfd4b`` - not part of our repo either, though.


ekpyron:
Git messes up the compound diff, but in the diffs of the individual commit it should be clear what is happening (just renaming and linking).
The test run passes...
Not sure we should do it like this, though, and what to do about 0.1.1 and 0.1.2 - or whether to care too much... e.g. we might just not do any of this for the old prereleases...

ekpyron:
Independently of this we might also want to check for which the commit hash in the file name actually matches what ``--version`` says - for most of them it should agree, but there may be some for which builds failed during the release process and it's actually wrong... might be worth investigating and solving similar to this here: we decide if the tag or what ``--version`` says is the "proper" name and create a symlink for the other...

PR #27, opened 2020-06-11
Release 0.6.10.

ekpyron:
This is what I get for comparison - with the binary downloaded again from the github release page...


ekpyron:
Wait, what? I think this now also has binaries that are cut short in the end... let me check...

ekpyron:
Now I think this one is fine...

PR #30, opened 2020-07-03
Github action for emscripten nightly builds

cameel:
Part of https://github.com/ethereum/solidity/issues/9258. Replaces #29.

Another go at the same problem. This time doing a full checkout and building `soljson.js` within the action instead of pulling it from CircleCI. This does make things simpler and pretty much solves all the outstanding problems from the previous PR. I just still need answers for the two questions below.

It's also significantly slower now: ~20 min for the build, ~3.5 min for git checkout, ~8 min for the update script.

I went with @ekpyron's suggestion and split the workflow into two jobs with `soljson.js` being passed between them as an artifact. This way we can still reuse the second part (that updates `solc-bin`) if we ever decide to go back to the solution with pulling from CircleCI.

#### Questions
- [x] ~I need an e-mail address to put in the committer field (currently it's `emscripten nightly action <solidity@example.com>`).~
    - I'll use builds@ethereum.org.
- [x] ~I have created a `master` branch in the repo and set it as the branch where the action adds new commits. Since there are external tools that rely on binaries hosted on GH pages, I think that it would be best to consider `gh-pages` frozen until we get S3 ready to serve the files instead and to work on `master` in the meantime.~
    - After discussing it on Gitter, I'm dropping the `master` branch. We'll keep updating `gh-pages` and just exclude files that go over the GH pages limit using jekyll config.

### DIFF
--- a/.github/workflows/nightly-emscripten.yml
+++ b/.github/workflows/nightly-emscripten.yml
@@ -1,0 +1,127 @@
+name: Emscripten nightly build + push to solc-bin
+
+on:
+  schedule:
+    # Run once a day, at midnight
+    - cron: '0 0 * * *'
+
+env:
+  TARGET_BRANCH: gh-pages
+  COMMITTER_NAME: emscripten nightly action
+  COMMITTER_EMAIL: builds@ethereum.org
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  build-emscripten-nightly:
+    runs-on: ubuntu-latest
+    outputs:
+      solidity-version: ${{ env.SOLIDITY_VERSION }}
+      nightly-version: ${{ env.NIGHTLY_VERSION }}
+      matching-nightlies-in-the-repo: ${{ env.MATCHING_NIGHTLIES_IN_THE_REPO }}
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          repository: 'ethereum/solidity'
+          ref: 'develop'
+          path: 'solidity/'
+
+      - name: Clone solc-bin repository without checking out a working copy
+        run: |
+          git clone --no-checkout --branch "$TARGET_BRANCH" "https://github.com/${GITHUB_REPOSITORY}.git" solc-bin/
+
+          # For some reason git stages all files for deletion when you use --no-checkout
+          cd solc-bin/
+          git reset HEAD --quiet
+
+      - name: Check if there's already a nightly with the same date or commit ID
+        run: |
+          cd solidity/
+          solidity_version=$("scripts/get_version.sh")
+          last_commit_timestamp=$(git log -1 --date=iso --format=%ad HEAD)
+          last_commit_date=$(date --date="$last_commit_timestamp" --utc +%Y.%-m.%-d)
+          last_commit_hash=$(git rev-parse --short=8 HEAD)
+
+          cd ../solc-bin/
+          matching_nightlies_in_the_repo="$(
+            git ls-files "bin/soljson-v${solidity_version}-nightly.${last_commit_date}+commit.*.js";
+            git ls-files "bin/soljson-v${solidity_version}-nightly.*+commit.${last_commit_hash}.js"
+          )"
+          nightly_version="v${solidity_version}-nightly.${last_commit_date}+commit.${last_commit_hash}"
+
+          echo "::set-env name=SOLIDITY_VERSION::${solidity_version}"
+          echo "::set-env name=NIGHTLY_VERSION::${nightly_version}"
+
+          # There's no way to just stop a job without failing and that would spam everyone with
+          # spurious e-mail notifications about the failure. Instead we have to make do with `if`s.
+          echo "::set-env name=MATCHING_NIGHTLIES_IN_THE_REPO::${matching_nightlies_in_the_repo}"
+
+      - name: Build soljson.js
+        if: "!env.MATCHING_NIGHTLIES_IN_THE_REPO"
+        run: |
+          cd solidity/
+          # Note that this script will spawn and build inside a docker image (which works just fine in github actions).
+          scripts/build_emscripten.sh
+
+      - name: Upload soljson.js as an artifact
+        if: "!env.MATCHING_NIGHTLIES_IN_THE_REPO"
+        uses: actions/upload-artifact@v2
+        with:
+          name: soljson.js
+          path: solidity/upload/soljson.js
+
+
+  test-emscripten-nightly:
+    runs-on: ubuntu-latest
+    needs: build-emscripten-nightly
+
+    env:
+      SOLIDITY_VERSION: ${{ needs.build-emscripten-nightly.outputs.solidity-version }}
+
+    if: "!needs.build-emscripten-nightly.outputs.matching-nightlies-in-the-repo"
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          repository: 'ethereum/solidity'
+
+      - name: Download soljson.js artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: soljson.js
+
+      - name: Run solc-js tests
+        run: |
+          test/externalTests/solc-js/solc-js.sh "${PWD}/soljson.js" "$SOLIDITY_VERSION"
+
+
+  add-nightly-and-push:
+    runs-on: ubuntu-latest
+    needs:
+      - build-emscripten-nightly
+      - test-emscripten-nightly
+
+    env:
+      NIGHTLY_VERSION: ${{ needs.build-emscripten-nightly.outputs.nightly-version }}
+
+    if: "!needs.build-emscripten-nightly.outputs.matching-nightlies-in-the-repo"
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ env.TARGET_BRANCH }}
+          path: 'solc-bin'
+
+      - name: Download soljson.js artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: soljson.js
+
+      - name: Set committer name and e-mail
+        run: |
+          cd solc-bin/
+          git config --local user.name "$COMMITTER_NAME"
+          git config --local user.email "$COMMITTER_EMAIL"
+
+      - name: Run add-nightly-and-push.sh
+        run: |
+          cd solc-bin/
+          ./add-nightly-and-push.sh ../soljson.js "$NIGHTLY_VERSION"
--- a/add-nightly-and-push.sh
+++ b/add-nightly-and-push.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+
+#------------------------------------------------------------------------------
+# Adds a nightly binary to a local checkout of solc-bin repository, updates
+# file lists, commits and pushes the changes to the remote repository.
+#
+# The script expects to find the working copy in its working directory with
+# the right branch already checked out and all the details necessary to make
+# a commit and push it (committer name, `origin` remote, etc.) configured.
+#
+# The script fails if the specified version is already present in the
+# repository.
+#------------------------------------------------------------------------------
+
+set -e
+
+die() {
+    echo
+    >&2 echo "ERROR: $@" && false
+}
+
+nightly_binary="$1"
+nightly_version="$2"
+(( $# == 2 )) || die "Expected exactly 2 parameters."
+
+if [[ $(git status --porcelain) != "" ]]; then
+    >&2 git status
+    die "Uncommitted and/or untracked files present. This script must be executed in a clean working copy of the repository."
+fi
+
+echo "Adding bin/soljson-${nightly_version}.js"
+[[ ! -e "bin/soljson-${nightly_version}.js" ]] || die "File bin/soljson-${nightly_version}.js already exists."
+cp "$nightly_binary" "bin/soljson-${nightly_version}.js"
+
+# Stage the new nightly before running the list update to be able to detect any unintended changes caused by the script.
+git add --verbose "bin/soljson-$nightly_version.js"
+
+npm config set package-lock false
+npm install
+npm run update
+rm -r node_modules/
+
+git add --verbose bin/soljson-nightly.js
+git add --verbose bin/list.*
+
+git commit --message "Nightly build ${nightly_version}"
+
+if [[ $(git status --porcelain) != "" ]]; then
+    >&2 git status
+    die "Unexpected file modifications found. This should never happen and might be the result of a bug."
+fi
+
+git push origin HEAD


.github/workflows/pull-soljson.yml
```diff
@@ -0,0 +1,118 @@
...
+          scripts/build_emscripten.sh
+
+      - name: Get nightly version
+        id: nightly-version
+        run: |
+          last_commit_timestamp=$(git log -1 --date=iso --format=%ad HEAD)
+          last_commit_date=$(date --date="$last_commit_timestamp" --utc +%Y.%-m.%-d)
+          last_commit_hash=$(git rev-parse --short=8 HEAD)
+          solidity_version=$(scripts/get_version.sh)
+          nightly_version="v${solidity_version}-nightly.${last_commit_date}+commit.${last_commit_hash}"
```
cameel: I will replace this with `get_nightly_version.sh` when https://github.com/ethereum/solidity/pull/9310 gets merged.

.github/workflows/pull-soljson.yml
```diff
@@ -0,0 +1,118 @@
...
+          path: upload/soljson.js
+
+  commit-soljson:
+    runs-on: ubuntu-latest
+    needs: build-emscripten
+
+    env:
+      COMMITTER_NAME: emscripten nightly action
+      # TODO: Set a proper e-mail address
+      COMMITTER_EMAIL: solidity@example.com
```
cameel: What would you suggest to use as an e-mail address here?

.github/workflows/nightly-emscripten.yml
```diff
@@ -0,0 +1,120 @@
...
+          path: upload/soljson.js
+
+  commit-soljson:
+    runs-on: ubuntu-latest
+    needs: build-emscripten
+
+    env:
+      COMMITTER_NAME: emscripten nightly action
+      # TODO: Set a proper e-mail address
+      COMMITTER_EMAIL: solidity@example.com
```
chriseth: We could use solidity@ethereum.org or builds@ethereum.org

chriseth: (the latter is an account already used for the ppa builds)

.github/workflows/nightly-emscripten.yml
```diff
@@ -0,0 +1,120 @@
...
+          ref: ${{ env.TARGET_BRANCH }}
+
+      - name: Make sure that the working copy is clean
+        run: |
+          echo "::debug::$(git status)"
+          test -z "$(git status --porcelain)" || exit 1
+
+      - name: Check if the nightly already exists
+        run: |
+          version_already_exists=$(git ls-files "bin/soljson-$NIGHTLY_VERSION.js" | sed q1 || echo true && true)
```
chriseth: Why not use `[ -e "file" ]`? This is a checkout with working copy, right?

.github/workflows/nightly-emscripten.yml
```diff
@@ -0,0 +1,120 @@
...
+          echo "::set-env name=VERSION_ALREADY_EXISTS::${version_already_exists}"
+
+      - name: Download soljson.js artifact
+        uses: actions/download-artifact@v2
+        if: "!env.VERSION_ALREADY_EXISTS"
```
chriseth: I find this all rather hard to read - is it possible to put basically everything in a shell script? Is there an advantage in splitting this into github action steps?

.github/workflows/nightly-emscripten.yml
ekpyron: Actually, I think we even *have* to use ``[ -e "file" ]`` here.
The ``git ls-files`` thing was meant to check if there is a uncommited new file in the repo, after copying it over, but we're now checking *before* copying over (which does make more sense), but then ``git ls-files`` like this here won't work.

.github/workflows/nightly-emscripten.yml
```diff
@@ -0,0 +1,120 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ env.TARGET_BRANCH }}
+
+      - name: Make sure that the working copy is clean
+        run: |
+          echo "::debug::$(git status)"
+          test -z "$(git status --porcelain)" || exit 1
```
ekpyron: Do we really need this? We just checked it out three lines up, didn't we? Or was this only for you testing stuff, then nevermind :-).

.github/workflows/nightly-emscripten.yml
ekpyron: Downloading the artifact here is github action specific, so it needs to stay like this... but maybe it'd already help to merge all of the stuff below to a single ``run:`` step? But I guess @cameel may have planned to do that anyhow and it's just split so far, because that makes it quicker to see what exactly failed in the github action output.

.github/workflows/nightly-emscripten.yml
cameel: I'm just being paranoid because of the file truncations :) This check is here to prove that files we're getting out of `git clone` are still clean. I treat it more like an assertion. Ideally, I'd put it at the end of the previous step but I don't think I can do that.

`scripts/travis-emscripten/publish_binary.sh` actually does something very similar. It runs `git clean` immediately after a fresh clone and checkout to make sure the tree is clean. It shouldn't be necessary unless Travis keeps files between runs (does it?):
```
git clone --depth 2 git@github.com:ethereum/solc-bin.git
cd solc-bin
git config user.name "travis"
git config user.email <email>
git checkout -B gh-pages origin/gh-pages
git clean -f -d -x
```

.github/workflows/nightly-emscripten.yml
cameel: @ekpyron Both `[ -e "file" ]` and `git ls-files` would work. Note that I dropped the `--other` option so I'm actually checking for committed files.

Both will have the same effect as long as there are no uncommitted files - and the previous step enforces that. But semantically the `git ls-files` check is closer to what we really mean. If, for whatever reason, the file happens to be on disk but not in the repo, it does not mean that the nightly has already been built and we should not skip the remaining steps.

As a bonus, `git ls-files` works with and without a checkout. One less thing to change if the repo starts growing way beyond the original 10 GB limit and at some point we decide to go back to a checkout-less version.

.github/workflows/pull-soljson.yml
ekpyron: It's merged ;-)!

.github/workflows/nightly-emscripten.yml
chriseth: Hm, the previous scripts stopped if there was already a version with the same date - maybe we should also be more strict here?

.github/workflows/nightly-emscripten.yml
cameel: The advantage is that you get more useful feedback when a job fails. E.g. it shows you the values of environment variables between steps. I think that's helpful not only when writing it but would also make diagnosing future failures easier. And any more complex operations are already in scripts. The rest are just sanity checks, and simple git or file operations so basically just glue between real work.

Also, that's a habit from Ansible :) GH action syntax is **very** similar to Ansible. It's a big YAML list with steps described by dicts, each with a `name` and extra properties defining the command. The main difference being that Ansible has lots of built-in commands and you can only use one per step, which forces you to do work in small steps. In GH actions most steps just contain raw shell input so grouping it into a script is actually possible.

Another reason is that I expected that most of it would be pretty specific to GH actions and not portable. But looks like it's not that bad and with some minimal assumptions that I can document, the commands would work on any other CI.

So I can put most of it into a single script if you prefer that. I can get some of the feedback back by just adding more `echo` statements. A Bash script might also be easier to review and could sit in the `solidity` repo.

.github/workflows/nightly-emscripten.yml
cameel: Sounds like a good idea. Though the check might break if we ever change the date format (and I wanted to propose just that ;))

.github/workflows/nightly-emscripten.yml
ekpyron: Fair enough - a little paranoia doesn't hurt I guess!

.github/workflows/nightly-emscripten.yml
chriseth: One of the reasons I think a script would be better is that in the current format, you always have to think about adding `        if: "!env.VERSION_ALREADY_EXISTS"` instead of just using a proper if statement  with a body...

ekpyron: If ``ls-files`` really doesn't need the solc-bin checkout, we could move that check to the end of the ``build_emscripten`` job above and then have the entire ``commit-soljson`` job only run if the version doesn't yet exist, i.e. have one such ``if`` at the very outer scope? Not sure that's that much better...

ekpyron: Or we could just make this job fail, if the version already exists...
Actually, how is this supposed to work? Right now this will keep creating nightlies with the same commit hash every day as long as there are no commits on develop, right? Then we can assume that it's a proper error if this script runs with the version already existing and *should* just make this job fail then - then we don't need ``if``s and are notified, if this happens.
The alternative would be to only build a "new" nightly, if it's actually new, i.e. when the resulting binary is new...

ekpyron: Ah right, that's basically what @chriseth is suggesting above. I'd also say, just let it fail, if it already exists, then we don't need any of this.

.github/workflows/pull-soljson.yml
cameel: Done.

.github/workflows/nightly-emscripten.yml
cameel: Done. Replaced with builds@ethereum.org.

cameel:
Changing target branch to `gh-pages`.

I have also rebased it on top of it. Still working on moving parts of the action into a script.

.circleci/config.yml
```diff
@@ -0,0 +1,8 @@
+version: 2.1
+
+jobs:
+  build:
+    docker:
+      - image: cimg/base
+    steps:
+      - run: echo "This is an empty config just to keep CircleCI quiet."
```
ekpyron: :-D

cameel: I got a bit tired of it complaining about missing config :P

I'll actually put it in a separate PR but I wasn't sure if it would even work or if it would just ignore it and keep failing so I decided to try it out here.

cameel: #34

.github/workflows/nightly-emscripten.yml
```diff
@@ -0,0 +1,96 @@
...
+          last_commit_timestamp=$(git log -1 --date=iso --format=%ad HEAD)
+          last_commit_date=$(date --date="$last_commit_timestamp" --utc +%Y.%-m.%-d)
+          last_commit_hash=$(git rev-parse --short=8 HEAD)
+
+          cd ../solc-bin/
+          matching_nightlies_in_the_repo="$(
+            git ls-files "bin/soljson-v${solidity_version}-nightly.${last_commit_date}+commit.*.js";
+            git ls-files "bin/soljson-v${solidity_version}-nightly.*+commit.${last_commit_hash}.js"
+          )"
+          nightly_version="v${solidity_version}-nightly.${last_commit_date}+commit.${last_commit_hash}"
```
cameel, line 50: I went back to constructing nightly version manually (rather than using `get_nightly_version.sh`) because now I need the commit hash and date anyway and I have all the information needed to do it at hand. The script no longer adds much value, and just introduces the possibility of things going out of sync.

Should I leave it like that (and perhaps remove `get_nightly_version.sh`) or should I add another script (say, `find_matching_nightlies $date $commit`) in `solidity` repo alongside it?

.github/workflows/nightly-emscripten.yml
```diff
@@ -0,0 +1,96 @@
...
+      - name: Set committer name and e-mail
+        run: |
+          cd solc-bin/
+          git config --local user.name "$COMMITTER_NAME"
+          git config --local user.email "$COMMITTER_EMAIL"
+
+      - name: Run add-nightly-and-push.sh
+        run: |
+          cd solc-bin/
+          ./add-nightly-and-push.sh ../soljson.js "$NIGHTLY_VERSION"
```
cameel, line 127: I can add this script to `solidity` repo but then I'd have to either clone the repo in this job just to get access to it or pass it from the previous job as an artifact. For now I put it in the root dir of `solc-bin`.

add-nightly-and-push.sh
```diff
@@ -0,0 +1,55 @@
...
+# Adds a nightly binary to a local checkout of solc-bin repository, updates
+# file lists, commits and pushes the changes to the remote repository.
+#
+# The script expects to find the working copy in its working directory with
+# the right branch already checked out and all the details necessary to make
+# a commit and push it (committer name, `origin` remote, etc.) configured.
+#
+# The script fails if the specified version is already present in the
+# repository.
+#------------------------------------------------------------------------------
```
cameel, line 13: Should I add the license boilerplate here? Other scripts in this repo don't have it. And it might need to be reworded slightly (not sure).

.github/workflows/nightly-emscripten.yml
cameel: Checks added.

.github/workflows/nightly-emscripten.yml
cameel: Done. Moved most of the job into a script and put the whole second job in an `if`.

cameel:
I think the PR is now ready for the final review and merge. I can't test it directly here so I set it up as a hourly job in my fork (https://github.com/cameel/solc-bin). We'll see if it keeps running without problems.

ekpyron:
The current test failure in this PR here, is something we have seen a lot and I still don't really understand.
There's some days with multiple "nightlies" that day - apparently depending on who runs the script the sorting according to the commit hashes is different... The CI seems to order inverse-lexicographically, descending letters first, then numbers... which is as good as any order and the same that I always got locally... but other systems seem to produce another order (which always seemed rather random to me)...

cameel:
@ekpyron I checked semver spec and looks like it does not specify ordering for metadata (the part after the plus sign):

[Semantic Versioning 2.0.0 > bullet point 10](https://semver.org/#spec-item-10)
> Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+exp.sha.5114f85, 1.0.0+21AF26D3—-117B344092BD.

So e.g. `0.4.2-nightly.2016.9.17+commit.212e0160` is treated as equal to `v0.4.2-nightly.2016.9.17+commit.a78e7794`. We have to add an extra commit hash comparison if we want the order to be stable.

But the weird part is - why do we have several nightlies from 2016-09-17? Deleting all but one would solve the problem too :)

chriseth:
There should only be one nightly per day.

ekpyron:
> There should only be one nightly per day.

Sure there should, but what do we do with the old ones for which this *isn't* the case :-)?

What's the most recent one for which there are more than one? If it's 0.4.2, as the CI run indicates, then we could just remove *all* nightlies older than 0.4.3 as well (I chose how many to already remove from solc-bin quite arbitrarily anyways, so little harm in removing a few more, is there?).

cameel:
> we could just remove all nightlies older than 0.4.3 as well

Do you mean permanently? The nightlies that were already removed are going to be restored soon and just excluded from jekyll config. So that's a bit different from removing the "duplicates".

ekpyron:
> > we could just remove all nightlies older than 0.4.3 as well
> 
> Do you mean permanently? The nightlies that were already removed are going to be restored soon and just excluded from jekyll config. So that's a bit different from removing the "duplicates".

Ok, that's a fair point then... But on what basis would you choose one to keep and delete others? It's not "duplicates" if it's different hashes - it's different versions we did publish once... what justifies keeping one and not another?

Also: if we exclude them from the jekyll config, then we also have to exclude them from the lists, don't we? It would be weird, if the lists talk about files that don't exist... and that would be equivalent to removing them entirely as far as this particular issue is concerned...

cameel:
Yeah, I know they're not identical. That's why I put it in quotes. I really meant that all but one are redundant. I don't think any one of them is more important than the others so it does not really matter which one we keep.

> Also: if we exclude them from the jekyll config, then we also have to exclude them from the lists, don't we?

Right. That's still an open question. I think I asked about it somewhere before (on Gitter?) but it got lost in the discussion.

The list is indeed a problem. I think we should remove the files from the list but then we'd have to somehow serve different lists from GH pages. And that's not the only problem. `soljson.js` is now a symlink and `soljson-latest.js` and `soljson-nightly.js` in `bin/` will be too (after #33). They will keep working because jekyll replaces the link with a copy but now that copy won't match the latest visible release/nightly.

Maybe we could somehow add a second set of lists and switch them over during rendering to GH pages. And change symlinks on the fly during rendring too - but it's getting complicated quickly. Freezing the branch seems like a much simpler solution so that's what I would suggest instead.

ekpyron:
Quote from gitter:
```
@ekpyron: @cameel @chriseth Maybe we shouldn't complicate ethereum/solc-bin#30 just because of ancient nightlies :-)... what's the best way to resolve that quickly? No-one will ever really need those old 0.4.2 nightlies...
@chriseth yes, just delete them
```
I agree with that - just get rid of those old nightlies altogether and the problem is solved. We also don't need to bring the ones already deleted back.
EDIT: at least for now, so that this doesn't stall things. If the need *ever* arrises, we can still think about it, but I think it highly unlikely that anyone will ever need a 0.4.2 nightly...

chriseth:
Clarification: I was talking about the duplicate nightlies as in https://github.com/ethereum/solc-bin/pull/30#issuecomment-656021871

aarlt (commented):
I was wondering whether the nightlies should not also get tested before publishing. There seem to be a script for testing emscripten `./scripts/test_emscripten.sh`. It also is referenced in `.travis.yml`.

cameel:
@aarlt Right. They probably should. I'll get back to this and add the check when I'm done with the other PRs.

ekpyron:
Yeah, I was wondering whether to suggest running emscripten tests as well.
We could, but one could argue either way. If we had fetched the nightlies from CircleCI we wouldn't have tested them either - and no-one should actually use the nightlies for any production use anyways.

It wouldn't be too hard in any case - we can just spawn a job in between the two existing ones - you should be able to pretty much take over the ``test_emscripten`` job from https://github.com/ekpyron/solidity/blob/5a9345798e396e79d18faed61ed1de5c094557ef/.github/workflows/emscripten.yml - you'd just need to change the checkout URL for the solidity checkout (and pass over the version, resp. re-execute the get_version script).
We could think about just having it in there (it failing should then send out a mail to us I think) - or whether to make it an actual requirement for publishing the nightlies.

cameel:
Thanks. I'll use that.

By the way, if you think this check is not all that essential then I can also do it in a separate PR not to block this one.

ekpyron:
Separate PR seems fine to me for it!

.github/workflows/nightly-emscripten.yml
```diff
@@ -0,0 +1,96 @@
...
+          echo "::set-env name=NIGHTLY_VERSION::${nightly_version}"
+          echo "::set-env name=MATCHING_NIGHTLIES_IN_THE_REPO::${matching_nightlies_in_the_repo}"
+
+      - name: Build soljson.js
+        if: "!env.MATCHING_NIGHTLIES_IN_THE_REPO"
+        run: |
+          cd solidity/
+          # Note that this script will spawn and build inside a docker image (which works just fine in github actions).
+          scripts/build_emscripten.sh
+
```
chriseth, line 65: Should we maybe run at least a simple test?

.github/workflows/nightly-emscripten.yml
chriseth: Somehow I feel that the whole version juggling here (especially if the version is already there) could be better placed inside the update script, or at least in a nodejs script that uses the same utilities.

.github/workflows/nightly-emscripten.yml
cameel: Done. I added third job that runs `solc-js` tests on the nightly, based on @ekpyron's GH actions.

.github/workflows/nightly-emscripten.yml
cameel: After giving it some thought, how about moving the regex matching from the `update` script into a separate one that would split binary name into components, e.g.:
```
$ ./split-release-binary-name.sh "soljson-v0.6.1-nightly.2019.12.20+commit.ece6463f.js"
```
```
{
    "name": "soljson",
    "version": "0.6.1",
    "type": "nightly",
    "timestamp": "2019.12.20",
    "commit": "ece6463f",
    "extension": "js"
}
```
And we could also replace `get_nightly_version.sh` with one that does the reverse - gets this dict on stdin and prints the name. If there's nothing on stdin it would get current values from `git` and `date`.

JSON output should make it easy to use `split-release-binary-name.sh` both in nodejs scripts and in the shell (with the help of `jq`).

Then, in this action, we could run `git ls-files` to get the list of all files in `bin/`, pass each name through the script and run the result through `jq` to filter out commits and dates that do not match. Could even be a one-liner, though we'd probably want to split it for readability.

chriseth: Hm, maybe it's perfectly fine as it is :)

.github/workflows/nightly-emscripten.yml
```diff
@@ -0,0 +1,124 @@
...
+          cd ../solc-bin/
+          matching_nightlies_in_the_repo="$(
+            git ls-files "bin/soljson-v${solidity_version}-nightly.${last_commit_date}+commit.*.js";
+            git ls-files "bin/soljson-v${solidity_version}-nightly.*+commit.${last_commit_hash}.js"
+          )"
+          nightly_version="v${solidity_version}-nightly.${last_commit_date}+commit.${last_commit_hash}"
+
+          echo "::set-env name=SOLIDITY_VERSION::${solidity_version}"
+          echo "::set-env name=NIGHTLY_VERSION::${nightly_version}"
+          echo "::set-env name=MATCHING_NIGHTLIES_IN_THE_REPO::${matching_nightlies_in_the_repo}"
```
chriseth, line 57: Can't we just fail the build if there is a matching nightly? This would also stop all subsequent steps from being executed.

cameel: We could but then we would all be getting an e-mail notification about the failed job. I think it's better not to desensitize people to real failures by having spurious ones by design :)

I wish there was some way to just stop the job but as far as I know currently there isn't. Though I've seen people complain about that on github's community forum so maybe it will change in the future.

.github/workflows/nightly-emscripten.yml
cameel: Maybe the description above makes it look more complex that it would actually be. But yeah, nothing beats the simplicity of just not adding more stuff so that's fine too :)

.github/workflows/nightly-emscripten.yml
chriseth: Ah I see! Can you document the reason for not failing?

cameel: Done.

chriseth (approved):

cameel:
So looks like this is ready to get merged.

I didn't get an answer about [license](https://github.com/ethereum/solc-bin/pull/30#discussion_r451724312) and [keeping the scripts in `solc-bin`](https://github.com/ethereum/solc-bin/pull/30#discussion_r451722626) so I assume it's OK as is.

PR #31, opened 2020-07-04
Update node.js dependencies to the latest versions to pull in newer keccak module

cameel:
This resolves the issue with `ethereumjs-util` pulling in an old version of `keccak` version that does not build on node.js >= 12. The problem was not making `npm install fail` because there's a fallback to the pure JS implementation but it ran slower because of that. After the update the script runs in ~5 minutes in a github action, compared to ~8 minutes before.

#### `npm install` output before this PR
```
> keccak@1.4.0 rebuild /home/runner/work/solc-bin/solc-bin/node_modules/keccak
> node-gyp rebuild

make: Entering directory '/home/runner/work/solc-bin/solc-bin/node_modules/keccak/build'
  CXX(target) Release/obj.target/keccak/src/addon.o
../src/addon.cc: In static member function ‘static Nan::NAN_METHOD_RETURN_TYPE KeccakWrapper::Initialize(Nan::NAN_METHOD_ARGS_TYPE)’:
../src/addon.cc:37:47: error: no matching function for call to ‘v8::Value::IntegerValue()’
     unsigned int rate = info[0]->IntegerValue();
                                               ^

(...)

make: *** [Release/obj.target/keccak/src/addon.o] Error 1
keccak.target.mk:129: recipe for target 'Release/obj.target/keccak/src/addon.o' failed
make: Leaving directory '/home/runner/work/solc-bin/solc-bin/node_modules/keccak/build'
gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:194:23)
gyp ERR! stack     at ChildProcess.emit (events.js:315:20)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:275:12)
gyp ERR! System Linux 5.3.0-1031-azure
gyp ERR! command "/usr/local/bin/node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /home/runner/work/solc-bin/solc-bin/node_modules/keccak
gyp ERR! node -v v12.18.1
gyp ERR! node-gyp -v v5.1.0
gyp ERR! not ok 
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! keccak@1.4.0 rebuild: `node-gyp rebuild`
npm ERR! Exit status 1
npm ERR! 
[...]

PR #33, opened 2020-07-06
Replace latest and nightly copies in bin/ with symlinks

cameel:
Related to #32 (and actually depends on that PR). Part of https://github.com/ethereum/solidity/issues/9258.

~It's also a PR to the `master` branch.~ Changed to `gh-pages`.

It takes the change a bit further by removing all copies of `soljson.js` from `bin/` and replacing them with symlinks. This requires more changes to the `update` script and it's just a refactor, not a fix so I'm putting it in a separate PR #32.

### DIFF
--- a/update
+++ b/update
@@ -8,8 +8,28 @@ const semver = require('semver')
 const ethUtil = require('ethereumjs-util')
 const swarmhash = require('swarmhash')
 
-// This script updates the index files list.js and list.txt in the bin directory,
-// as well as the soljson-latest.js files.
+// This script updates the index files list.js and list.txt in the directories containing binaries,
+// as well as the 'latest' and 'nightly' symlinks.
+
+function updateSymlink (linkPathRelativeToRoot, targetRelativeToLink) {
+  const absoluteLinkPath = path.join(__dirname, linkPathRelativeToRoot)
+
+  fs.readlink(absoluteLinkPath, (err, linkString) => {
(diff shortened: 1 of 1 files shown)


cameel:
This triggers some really weird linter warnings:
```
(node:6978) Warning: Accessing non-existent property 'cat' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
(node:6978) Warning: Accessing non-existent property 'cd' of module exports inside circular dependency
(node:6978) Warning: Accessing non-existent property 'chmod' of module exports inside circular dependency
...
```
I'm not sure what they're all about but I wanted to finish for today so I'll investigate them tomorrow. Looks like putting it in a separate PR was a good idea.

cameel:
Looks like the `Accessing non-existent property` warnings were not introduced in this PR and they don't affect the checks anyway. It's actually a cosmetic problem that comes from `shelljs` which is a dependency of `eslint`, which in turn is a dependency of `standard`, which we use for linting. It has already been fixed upstream (https://github.com/shelljs/shelljs/issues/991) and the warning will go away after we update our dependencies (#31).

The checks were failing because were actual linter warnings below the warnings that I missed. Now they're fixed.

cameel:
Changing the target branch to `gh-pages`.

chriseth (approved):

PR #34, opened 2020-07-07
Add an empty CircleCI config to silence the empty job that fails on every PR

cameel:
Maybe we could just disable it in settings but I guess we would have done that long ago if that was that easy.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,8 @@
+version: 2.1
+
+jobs:
+  build:
+    docker:
+      - image: alpine
+    steps:
+      - run: echo "This is an empty config just to keep CircleCI quiet."


.circleci/config.yml
```diff
@@ -0,0 +1,8 @@
+version: 2.1
+
+jobs:
+  build:
+    docker:
+      - image: alpine
+    steps:
+      - run: echo "This is an empty config just to keep CircleCI quiet."
```
ekpyron, line 8: Does this actually need to be a whole job or does an empty file suffice? Also: can we instead just prevent CircleCI from starting to run on PRs here with some settings change?
I'm fine with merging this - but I'm not sure we really need it either :-).

cameel: Haven't tried an empty one. I tried empty `step` list and it failed JSON validations.

And yeah, disabling it in settings would be better if possible, but I don't have access to settings and I didn't want to bother @chriseth. I suspect that there's a good reason why it's not already disabled anyway.

cameel: I just tried an empty config and it behaves as if the file was missing.

If I leave only `version` in the file, it fails validation:
```
# Error calling workflow: 'workflow'
# Cannot find a definition for job named build
```

If I remove the `docker` key from the job I get
```
# ERROR IN CONFIG FILE:
# [#/jobs/build] 0 subschemas matched instead of one
# 1. [#/jobs/build] only 1 subschema matches out of 2
# |   1. [#/jobs/build] no subschema matched out of the total 2 subschemas
# |   |   1. [#/jobs/build] 0 subschemas matched instead of one
# |   |   |   1. [#/jobs/build] required key [docker] not found
# |   |   |   2. [#/jobs/build] required key [machine] not found
# |   |   |   3. [#/jobs/build] required key [macos] not found
# |   |   2. [#/jobs/build] required key [executor] not found
# |   |   |   A job must have one of `docker`, `machine`, `macos` or `executor` (which can provide docker/machine/macos information).
# 2. [#/jobs/build] expected type: String, found: Mapping
# |   Job may be a string reference to another job
```

Looks the one I came up with is as slim as it gets.

chriseth:
I turned off the notification - could not find a switch to turn of building...

cameel:
> I turned off the notification - could not find a switch to turn of building...

The funny thing is that I no longer see it listed as check in PRs but I'm still getting e-mail notifications :)

chriseth:
Yes, they are still running, but they do not post notifications to github. Actually circleci says that it runs the jobs as long as there is at least one user following the project - so maybe the solution is for you to unfollow

cameel:
I can unfollow or even just keep ignoring those e-mails but if there's no other way to actually disable it then maybe this no-op job is actually a better workaround? It would silence it for everyone, not just me.

PR #35, opened 2020-07-09
Release download script

cameel:

The script downloads latest release binaries for Linux, Mac OS and wasm from CircleCI. Files are placed in the following directory structure:
- `emscripten-asmjs/solc-emscripten-asmjs-<version>.js`
    - Files symlinked to `bin/`
- `emscripten-wasm32/solc-emscripten-wasm32-<version>.js`
    - Files symlinked to `wasm/`
- `linux-amd64/solc-linux-amd64-<version>`
- `macosx-amd64/solc-macosx-amd64-<version>`
- `windows-amd64/solc-windows-amd64-<version>.zip`

TODO:
- [x] Files do not have the executable bit set. **Done in #41**
- [x] ~Make sure `b_ubu` is the right source for the Linux build.~
    - We don't have static linux builds on CircleCI
- [x] Add `latest` symlinks for Linux and Mac OS and make the `update` script update them. **Done in #40**
- [x] Query github API only once rather than once per downloaded binary.
- [x] Make the script actually stop and display a sensible error message when it runs into Github API rate limiting (currently it prints `curl: (3) bad range specification in URL position 2:` and continues).
- [x] ~Don't overwrite binaries if they already exist (and maybe add a flag to force overwrite).~
    - Added information that they will be overwritten instead. They're supposed to be versioned anyway.

### DIFF
--- a/download-release.sh
+++ b/download-release.sh
@@ -3,7 +3,7 @@
 set -e
 set -o pipefail
 
-default_source=github
+default_source=circleci
 default_solidity_version=latest
 
 if [[ $1 == --help ]]; then
@@ -17,11 +17,11 @@ if [[ $1 == --help ]]; then
     echo "    ./$(basename "$0") --help"
     echo "    ./$(basename "$0") [source] [solidity_version] [solc_bin_dir]"
     echo
-    echo "    source           The source to get binaries from. Must be 'github'."
-    echo "                     Other sources may be added in the future."
+    echo "    source           The source to get binaries from. Can be 'circleci' or 'github'."
     echo "                     Default: '${default_source}'."
     echo "    solidity_version Version tag representing the release to download, including"
     echo "                     the leading 'v'. Use 'latest' to get the most recent release."
+    echo "                     In case of CircleCI 'latest' is the only supported value."
     echo "                     Default: '${default_solidity_version}'."
     echo "    solc_bin_dir     Location of the solc-bin checkout."
     echo "                     Default: current working directory."
@@ -29,7 +29,7 @@ if [[ $1 == --help ]]; then
     echo
     echo "Examples:"
     echo "    ./$(basename "$0") --help"
-    echo "    ./$(basename "$0") github latest"
+    echo "    ./$(basename "$0") circleci latest"
     echo "    ./$(basename "$0") github v0.6.9"
     echo "    ./$(basename "$0") github latest ~/solc-bin/"
     exit 0
@@ -66,10 +66,38 @@ expect_single_line() {
 }
 
 
+# JOB INFO FROM CIRCLECI API
+
+filter_jobs_by_name() {
+    local job_name="$1"
+    jq '[ .[] | select (.workflows.job_name == "'"$job_name"'") ]'
+}
+
+select_latest_job_info() {
+    # ASSUMPTION: Newer jobs always have higher build_num
+    jq 'sort_by(.build_num) | last'
+}
+
+
+# ARTIFACT INFO FROM CIRCLECI API
+
+filter_artifacts_by_name()  {
+    local artifact_name="$1"
+
+    jq '[ .[] | select (.path == "'"$artifact_name"'") ]'
+}
+
+
 # TAG INFO FROM GITHUB API
 
 filter_only_version_tags()  {
     jq '.[] | select (.name | startswith("v"))'
+}
+
+filter_tags_by_commit()  {
+    local commit_hash="$1"
+
+    jq '[ . | select (.commit.sha == "'"$commit_hash"'") ]'
 }
 
 filter_tags_by_name()  {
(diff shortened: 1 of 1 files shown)


cameel:
Script updated:
- Now it can also download releases from github.
- Added `--help`.
- List of tags is now fetched from github only once.
- Script now properly exits on errors. It wasn't working because `local` is a shell command by itself and its return value shadows the one from `$(...)`. Unfortunately fixing this requires separate declaration and assignment (`local a; a=$(...)`) or using only global variables.
- Regex in the `update` script is now correct.

update
```diff
@@ -11,7 +11,7 @@ const swarmhash = require('swarmhash')
 // This script updates the index files list.js and list.txt in the bin directory,
 // as well as the soljson-latest.js files.
 
-var dirs = ['/bin', '/wasm']
+var dirs = ['/macos', '/linux', '/wasm', '/bin']
```
chriseth: this does not have windows because it does not have a circelci config, right?

chriseth:
Should we include an architecture for linux, macos and windows?

update
cameel: Yes, exactly. I wanted to take care of windows builds in a separate PR, when I can actually get them.

cameel:
Good point, I think we should. We may want to support Mac OS on both arm and x86 at some point.

I'd put that both in the file name and in the directory name. I don't have a strong opinion on how granular the directories should be - other than that they should not contain too many entries - so for now I'm going with the directories representing the things that are forced on you (system, platform, etc.) and slight variations that are up to your preference (i.e. static vs dynamic, gcc vs clang) would be kept in the same directory and just have different file names.

cameel:
To summarize the discusion from Gitter, this is the naming convention I'm going to use:

```
emscripten-asmjs/solc-emscripten-asmjs-v0.5.0+commit.1d4f565a.js
emscripten-wasm32/solc-emscripten-wasm32-v0.5.0+commit.1d4f565a
linux-amd64/solc-linux-amd64-v0.5.0+commit.1d4f565a
macosx-amd64/solc-macosx-amd64-v0.5.0+commit.1d4f565a
windows-amd64/solc-windows-amd64-v0.5.0+commit.1d4f565a.zip
```

chriseth:
As said on gitter, I think we should start with the os instead of the architecture. Are  we really using wasm32 instead of 64?

Otherwise this looks good!

cameel:
Right. I updated the comment.

> Are we really using wasm32 instead of 64?

I'm not 100% sure but I think that's the case. @ekpyron confirmed it when I asked. I did a quick search earlier and I got the impression that wasm64 is still just on the roadmap. But maybe that was just about some specific implementation? Anyway, I'll check out build options to make sure.

ekpyron:
It's wasm32 - I think that's the only option that really exists today and is actually usable in browsers.

cameel:
@chriseth I'm trying to figure out if `b_ubu` is really the job I should use for the linux build and it looks like the size is wildly different from the `solc-linux-static` binary posted on the release page:
- `b_ubu`: 36.7 MB
- `solc-linux-static`: 9.0 MB

None of the other jobs matches either:
- `b_ubu_release`: 36.7 MB (and byte-for-byte identical to `b_ubu`)
- `b_ubu_clang`: 36.9 MB
- `b_ubu18`: 36.2 MB
- `b_archlinux`: 11.7 MB
- `b_ubu_cxx20`: no artifacts

See [CircleCI job for v0.6.11](https://app.circleci.com/pipelines/github/ethereum/solidity/6197/workflows/e0d9b26d-d706-4e16-a243-8e7762c9e96c)

`soljson.js` from `b_ems` job is also different, but only slightly: 22 629 520 bytes (github) vs 22 630 292 (circleci). Wasn't it supposed to be deterministic?

Surprisingly, the Mac OS build is identical byte-for-byte.

ekpyron:
@cameel 0.6.11 is the only exception there - the scripts used different emscripten versions for that one, but usually that won't be the case.
EDIT: Also as of now I don't think there are any static linux binaries built on CircleCI that can be used here...

cameel:
If CircleCI linux build is not static then why is it larger than the static one from Travis?

ekpyron:
While the CircleCI version is dynamically linked, it contains a static version of Z3, which is quite large - that's not in the "travis" build (although arguably it should be in the static release binary).

cameel:
ok. In that case it seems that everything is in order. I'll use `b_ubu`.

ekpyron:
You can't use that - that's dynamically linked :-).

cameel:
Oh, right, sorry. In that case: do we add a static build to CircleCI? An alternative would be to use one of @ekpyron's GH actions and fetch the binary from there (this would also ease the transition to fully automating it later).

ekpyron:
I think we should postpone this whole thing altogether for the time being - @chriseth also said on the gitter channel, that we don't really need this script for now, but rather need a directory structure, index files and hosting for the binaries (and we can just keep fetching the actual binaries manually from circleci, travis and appveyor, respectively from the release page the latter two will keep pushing to for now)

cameel:
I'm not sure if this will still be useful for anything so I'm leaving it as draft for now. It does work for MacOS and emscripten builds but obviously can't get static linux and windows ones because we don't have any.

I salvaged the github part and put it into #41. I added some modifications there: downloading windows release, not running the update script, asm.js/wasm detection, etc.

chriseth:
Maybe we can just close this.

cameel:
The script here is now a bit out of date (our artifacts changed, CircleCI has released API v2.0, etc.).

I'm going to close this in favor of https://github.com/ethereum/solidity/pull/12181, which reuses parts of this script.

PR #39, opened 2020-07-17
asm.js binaries in bin/ and wasm binaries in wasm/

cameel:
Depends on on #38. Part of https://github.com/ethereum/solidity/issues/9258.

Currently `bin/` contains a mix of wasm and asm.js binaries. This makes it hard to determine where an older release downloaded from github would go. This PR moves all wasm binaries from `bin/` to `wasm/` and adds symlinks in their place.

This required changing the `update` script to make it not ignore the newly added symlinks.

### DIFF
--- a/update
+++ b/update
@@ -55,7 +55,13 @@ dirs.forEach(function (dir) {
 
     // ascending list (oldest version first)
     const parsedList = files
-      .filter(function (file) { return dir === '/wasm' || !file.isSymbolicLink() })
+      .filter(function (file) {
+        // Skip symbolic links with less then 8 characters in the commit hash.
+        // They exist only for backwards-compatibilty and should not be on the list.
+        return dir !== '/bin' ||
+          !file.isSymbolicLink() ||
+          file.name.match(/^.+\+commit\.[0-9a-f]{8,}\.js$/)
+      })
       .map(function (file) { return file.name })
       .map(function (file) { return file.match(/^soljson-v([0-9.]*)(-([^+]*))?(\+(.*))?.js$/) })
       .filter(function (version) { return version })


cameel:
Here's the script I used to separate the files:
``` bash
#!/usr/bin/env bash

set -e

is_wasm() {
    # Just a heuristic but it does match the expected release versions in solc-bin so far.
    grep --fixed-strings --silent 'var wasmBinaryFile="data:application/octet-stream;base64,AGFzbQEA' "$@" && return 0
    return 1
}

for src_dir in bin wasm; do
    cd "$src_dir"
    for binary_name in $(find * -regex 'soljson-v[0-9.]+\+commit\..*\.js' -type f | sort); do
        full_version=$(echo "$binary_name" | sed -r 's/soljson-(v[0-9.]+\+commit\..*)\.js/\1/')

        if is_wasm "$binary_name"; then
            if [[ $src_dir == bin ]]; then
                echo "Moving wasm binary bin/${binary_name} to wasm/"
                dst_dir="wasm"
            else
                continue
            fi
        else
            if [[ $src_dir == wasm ]]; then
                echo "Moving asm.js binary wasm/${binary_name} to bin/"
                dst_dir="bin"
            else
                continue
            fi
        fi

        if [[ -e "../${dst_dir}/${binary_name}" ]]; then
            [[ -L "../${dst_dir}/${binary_name}" ]] || { echo "ERROR: File ${dst_dir}/${binary_name} already exists and is not a symbolic link"; exit 1; }
            rm "../${dst_dir}/${binary_name}"
        fi

        mv "$binary_name" "../${dst_dir}/${binary_name}"
        ln --symbolic "../${dst_dir}/${binary_name}" "$binary_name"
    done
    cd ..
done
```

chriseth (approved):

PR #40, opened 2020-07-17
Emscripten releases

cameel:
Depends on on #39. Part of https://github.com/ethereum/solidity/issues/9258.

This PR puts those emscripten releases that already exist in the repo into a new directory structure:
- I created symlinks from the new directories to `bin/` and `wasm/` and not the other way around because otherwise excluding the new directories from Jekyll config would result in errors (link targets would not exist in the rendered version).
- There are no `soljson` versions on github release pages that would not already be in `solc-bin` - so this PR just adds links.
- In two cases files in the repo and on the release pages differ:
    1. asm.js builds of `v0.2.0`, `v0.2.1` and `v0.4.11` are not byte-for-byte identical and the sizes differ slightly.
    2. asm.js and wasm builds of `v0.4.11`, `v0.5.9` and `v0.5.11` come from a different commit than the ones on the release page (as reported in https://github.com/ethereum/solidity/issues/7512).


cameel:
Here's the one-off script I used to create symlinks and verify that each directory contains the right files:
``` bash
#!/usr/bin/env bash

set -e

is_wasm() {
    # Just a heuristic but it does match the expected release versions in solc-bin so far.
    grep --fixed-strings --silent 'var wasmBinaryFile="data:application/octet-stream;base64,AGFzbQEA' "$@" && return 0
    return 1
}

for bin_dir in bin wasm; do
    cd "$bin_dir"
    for binary_name in $(find * -regex 'soljson-v[0-9.]+\+commit\..*\.js' -type f | sort); do
        full_version=$(echo "$binary_name" | sed -r 's/soljson-(v[0-9.]+\+commit\..*)\.js/\1/')

        if [[ $bin_dir == wasm ]]; then
            is_wasm "$binary_name" || { echo "ERROR: Found non-wasm binary in wasm/: ${bin_dir}/${binary_name}"; exit 1; }
            link_location="emscripten-wasm32/solc-emscripten-wasm32-${full_version}.js"
        else
            ! is_wasm "$binary_name" || { echo "ERROR: Found non-asm.js binary in bin/: ${bin_dir}/${binary_name}"; exit 1; }
            link_location="emscripten-asmjs/solc-emscripten-asmjs-${full_version}.js"
        fi

        mkdir -p "$(dirname "../$link_location")"
        ln --symbolic "../${bin_dir}/${binary_name}" "../$link_location"
        printf "%-40s -> %s\n" "$link_location" "${bin_dir}/${binary_name}"
    done
    cd ..
done
```

chriseth (approved):

PR #41, opened 2020-07-17
Script for downloading release binaries from github

cameel:
Depends on on #39. Part of https://github.com/ethereum/solidity/issues/9258.

This is the part of #35 that only deals with github. I used the script to get old releases from github and I'm also going to hook it up as a PR check to compare newly added releases with those on the release page.

The script:
- Detects wasm/asm.js `soljson` and puts it in the right directory.
- Sets the executable flag for linux and macos binaries.
- Uses regexes to properly recognize artifacts with versions or other quirks in names.
- Github does not provide commit hashes for individual releases and tags so the script must fetch the whole list of tags to get it. Currently the list still fits on one page but I added pagination to future-proof it.
- Only downloads binaries. The `update` script needs to be run separately (that was simpler than adding a flag to disable the update).
- The only old release it won't handle is the gzipped `soljson.js` from v0.1.3. I could have added a special case for it but decided that it was not worth the effort. I downloaded that release manually.

### DIFF
--- a/download-release.sh
+++ b/download-release.sh
@@ -1,0 +1,287 @@
+#!/usr/bin/env bash
+
+set -e
+set -o pipefail
+
+default_source=github
+default_solidity_version=latest
+
+if [[ $1 == --help ]]; then
+    echo "Downloads release binaries, puts them at the right locations in a local"
+    echo "checkout of the solc-bin repository and updates the file lists."
+    echo
+    echo "WARNING: The binaries will be overwritten if they already exist."
+    echo
+    echo
+    echo "Usage:"
+    echo "    ./$(basename "$0") --help"
+    echo "    ./$(basename "$0") [source] [solidity_version] [solc_bin_dir]"
+    echo
+    echo "    source           The source to get binaries from. Must be 'github'."
+    echo "                     Other sources may be added in the future."
+    echo "                     Default: '${default_source}'."
+    echo "    solidity_version Version tag representing the release to download, including"
+    echo "                     the leading 'v'. Use 'latest' to get the most recent release."
+    echo "                     Default: '${default_solidity_version}'."
+    echo "    solc_bin_dir     Location of the solc-bin checkout."
+    echo "                     Default: current working directory."
+    echo
+    echo
+    echo "Examples:"
+    echo "    ./$(basename "$0") --help"
+    echo "    ./$(basename "$0") github latest"
+    echo "    ./$(basename "$0") github v0.6.9"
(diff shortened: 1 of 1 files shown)


cameel:
Moving back to draft since after the partial revert of #39 it does require some tweaks.

download-release.sh
```diff
@@ -0,0 +1,287 @@
+#!/usr/bin/env bash
```
axic, line 1: Random comment: do you want to enable `shellcheck` on this repo prior to this commit?

cameel: I just ran into this comment randomly, sorry for not replying earlier.

This is actually a good reason to move scripts I originally put into this repo to the main one. I'm planning to do that eventually. I already did that with bytecode PR check.

Not sure about this particular script though. I'd like to salvage it if possible but currently I'm not sure if we actually have a use for it. We should probably talk about it on one of the planning meetings but this currently isn't too high on the priority list for me.

axic: Not sure it makes sense moving the `s3-sync.sh` to another repo, so perhaps better to have shellcheck here in any case.

cameel: Well, I think that even `s3-sync.sh` would fit right in with other scripts that are already in https://github.com/ethereum/solidity/tree/develop/scripts/solc-bin. I even think it might make sense to move `update` there. `solc-bin` with it's current size is not very convenient to use for coding and after doing that for some time I'm more and more convinced that it should contain only binaries and nothing else.

One downside of having code in this repo is that accessing it requires getting a checkout, which is big and wastes a lot of time. This is sometimes an obstacle when you want to have multliple jobs in a single workflow and not all of them need to access binaries. It's not that hard to work around - you can pass the scripts as artifacts or download individual files via github API - but it's still a bit annoying when you need to do this.

To be honest I'm not strongly convinced that we **must** have everything in the main repo - I see both upsides and downsides - I'm just currently leaning more towards that option. Having shellcheck here would not be bad either.

cameel: Also, it would be nice to have some of these scripts rewritten in Python eventually and for that we already have nice linting and testing infrastructure in main repo. For example the bytecode comparison check for PRs is written in Bash and this is the main reason why I gave up on running it on Windows binaries (they're being ignored currently). Originally I wanted to support them but I realized that I'd have to write a separate PowerShell or CMD version of the scripts just for that one platform.

cameel:
The script here is now a bit out of date (our artifacts changed, CircleCI has released API v2.0, etc.).

I'm going to close this in favor of https://github.com/ethereum/solidity/pull/12181, which reuses parts of this script.

PR #43, opened 2020-07-17
Revert "asm.js binaries in bin/ and wasm binaries in wasm/"

cameel:
Part of https://github.com/ethereum/solidity/issues/9258.

This PR reverts #39, which inadvertently broke `solc-js` which is downloading stuff directly from the repo (rather than from GH pages) and due to that can't handle symlinks.

> There's also this in other failing jobs (e.g. https://app.circleci.com/pipelines/github/ethereum/solidity/6672/workflows/79fb8b48-0558-4cf9-b3da-aeae56a83edc/jobs/392350):
> ```
> Downloading version soljson-v0.6.11+commit.5ef660b1.js
>  
> Hash mismatch: 0xf0abd02c495a0b4c5c9a7ff20de8b932e11fc3066d0b754422035ecd96fcdbbc vs 0xac282607f7beba008c6fb82e5c453b21d9b9d8d28cddd465c150cc27cae4f1ca
> ```

I'm not reverting the changes to the `update` script since they make the script work for symlinks in other recently added directories.

PR #44, opened 2020-07-17
Convert bin/soljson-latest.js back from symlink to a copy

cameel:
Turns out that the "Unexpected number" error is still there even after #43. I just got it while running `solc-js` tests locally:
```
# loading remote version - development snapshot
soljson-latest.js:1
soljson-v0.6.11+commit.5ef660b1.js
          ^^

SyntaxError: Unexpected number
    at wrapSafe (internal/modules/cjs/loader.js:1117:16)
    at Module._compile (internal/modules/cjs/loader.js:1165:27)
    at requireFromString (/tmp/tmp.MHw54T6wnQ/solc/node_modules/require-from-string/index.js:28:4)
    at IncomingMessage.<anonymous> (/tmp/tmp.MHw54T6wnQ/solc/wrapper.js:338:35)
    at IncomingMessage.emit (events.js:326:22)
    at endReadableNT (_stream_readable.js:1226:12)
    at processTicksAndRejections (internal/process/task_queues.js:80:21)
npm ERR! Test failed.  See above for more details.
```

It's a different (but related) problem. This test uses `solc-js` to download `bin/soljson-latest.js`. It's a symlink now and `soljson-v0.6.11+commit.5ef660b1.js` in the above is actually the file content (rather than a name) which `solc-js` tries to intepret as JavaScript and fails.

In this PR I'm reverting the change that made this file a symlink and changing the `update` script back to handle this. While at it, I added safeguards against overwriting symlink targets to prevent bugs like #32 from popping up again. I'm keeping `bin/soljson-nightly.js` as a symlink since tests don't rely on it and really, `solc-js` will have this problem with any of the symlinks in `bin/` of which there are many - this is just a workaround for them, not a fix.

### DIFF
--- a/update
+++ b/update
@@ -31,6 +31,33 @@ function updateSymlink (linkPathRelativeToRoot, targetRelativeToLink) {
   })
 }
 
+function updateCopy (srcRelativeToRoot, destRelativeToRoot) {
+  fs.readFile(path.join(__dirname, srcRelativeToRoot), function (err, data) {
+    if (err) {
+      throw err
+    }
+
+    const absoluteDest = path.join(__dirname, destRelativeToRoot)
+    fs.stat(absoluteDest, function (err, stats) {
+      if (err && err.code !== 'ENOENT') {
+        throw err
+      }
+
+      // If the target is a symlink, we want to replace it with a copy rather than overwrite the file it links to
+      if (!err && stats.isSymbolicLink()) {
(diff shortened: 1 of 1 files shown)


chriseth (approved):

cameel:
Oh, there were some of those pesky linting errors again... Should be fixed now.

PR #46, opened 2020-07-22
Add missing wasm symlinks to release 0.6.12

cameel:
Fixes #45.

### DIFF
--- a/emscripten-wasm32/list.js
+++ b/emscripten-wasm32/list.js
@@ -1,5 +1,6 @@
 
 var soljsonSources = [
+  "solc-emscripten-wasm32-v0.6.12+commit.27d51765.js",
   "solc-emscripten-wasm32-v0.6.11+commit.5ef660b1.js",
   "solc-emscripten-wasm32-v0.6.10+commit.00c0fcaf.js",
   "solc-emscripten-wasm32-v0.6.9+commit.3e3065ac.js",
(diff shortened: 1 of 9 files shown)


chriseth (approved):

chriseth:
travis job failed because of missing newline...

cameel:
What? In a symlink? WTF.

cameel:
Wait, it's probably just that the diff after `update` is not clean. I'll rerun it locally and see what might be the cause.

cameel:
Looks like the `update` script did not update the `latest` symlink in `wasm/` on my machine. Not sure why - I ran the clean version from `gh-pages` which should have done it. Anyway, rerunning locally to fix it.

cameel:
Fixed.

I think I know what's going on. It did not wait for `wasm/` to be processed (which includes creating symlinks to the new release in `emscripten-wasm32/`) before it started processing `emscripten-wasm32/`. It did not see the new symlink and therefore did not update the lists. That's despite doing the second update in a callback which I added specifically to prevent this.

This is the same issue I've just been fighting with after modifying the `update` script. This means that my changes are fine and the issue is actually already there, just not always reproducible.  I'll debug it tomorrow - for now, just rerunning it should be enough to fix this PR.

chriseth (approved):

cameel:
I just pushed #49 which automates creation of these symlinks.

PR #47, opened 2020-07-23
Github action for mirroring repository content to an S3 bucket

cameel:

The action is configured to run on every push to `gh-pages` and use AWS CLI tool to sync modified files (which is determined by timestamps and file sizes).

### Secrets
This action requires two secrets to be added in repository settings:
- `AWS_ACCESS_KEY_ID`
- `AWS_SECRET_ACCESS_KEY`

To get them:
- Log into the Solidity organization on AWS
- Click the "Management console" link next to the "AdministratorAccess" role
- Go to [All Services > IAM > Users > `solc-bin-mirror-github-action` > Security credentials](https://console.aws.amazon.com/iam/home?region=eu-central-1#/users/solc-bin-mirror-github-action?section=security_credentials).
- Click "Create access key button".

It will show you the key ID and secret. You can't see them again once you close the dialog - in that case just generate a new pair and delete the old one.

I created this user specifically for this action and its privileges are limited to reading/writing S3 buckets (note: all buckets, not just `solc-bin`). It can't access any other AWS services.

### DIFF
--- a/.github/workflows/s3-mirror.yml
+++ b/.github/workflows/s3-mirror.yml
@@ -1,0 +1,41 @@
+name: Mirror repository content to an S3 bucket
+
+on:
+  push:
+    branches:
+      - gh-pages
+
+  schedule:
+    # Pushes to `gh-pages` done from other actions cannot trigger this action so we also want it to run
+    # on a schedule. Let's give the nightly job a 1h head-start and run every day at 1:00.
+    - cron: '0 1 * * *'
+
+env:
+  S3_BUCKET: solc-bin
+  S3_REGION: eu-central-1
+
+jobs:
+  push-to-s3:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+
+    steps:
+      - name: Wait for other instances of this workflow to finish
+        # It's not safe to run two S3 sync operations concurrently with different files
+        uses: softprops/turnstyle@v1
+        with:
+          same-branch-only: no
+
+      - name: Configure the S3 client
+        run: |
+          aws configure set default.region "$S3_REGION"
+          aws configure set aws_access_key_id '${{ secrets.AWS_ACCESS_KEY_ID }}'
+          aws configure set aws_secret_access_key '${{ secrets.AWS_SECRET_ACCESS_KEY }}'
+
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Sync the S3 bucket
+        run: |
+          ./sync-s3.sh "$S3_BUCKET"
--- a/sync-s3.sh
+++ b/sync-s3.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+
+#------------------------------------------------------------------------------
+# Uploads the content of the local working copy to a storage bucket on
+# Amazon S3. Removes any files that do not exist locally. Files in the root
+# directory with names starting with a dot or an underscore are not uploaded.
+#
+# The script assumes that the AWS CLI tool is installed and already configured
+# with credentials allowing it to modify the bucket.
+#
+# NOTE: There's no built-in mechanism for updating an S3 bucket in an atomic
+# way. Only individual file updates are atomic. This means that during the
+# sync clients will see the intermediate state with some files missing or not
+# yet updated. Since the binaries are never modified or removed from the repository
+# under normal circumstances, updating file lists last is enough to alleviate this.
+#
+# When running multiple instances of this script concurrently on different
+# revisions it's theoretically possible to end up with any combination of
+# their files in the bucket so it should be avoided.
+#
+# WARNING: The script destructively modifies the working copy. Always run it
+# on a fresh clone!
+#------------------------------------------------------------------------------
+
+set -eo pipefail
+
+die() { >&2 echo "ERROR: $@" && false; }
+
+s3_bucket="$1"
+(( $# == 1 )) || die "Expected exactly 1 parameter."
+
+[[ $(git rev-parse --is-shallow-repository) == false ]] || die "This script requires access to full git history to be able to set file timestamps correctly."
+
+echo "===> Updating file modification timestamps to match commits"
+# NOTE: `aws s3 sync` compares file timestamp and size to decide whether to upload it or not.
+readarray -t files < <(git ls-files)
+for file in "${files[@]}"; do
+    full_time="$(git log --max-count 1 --pretty=format:%cd --date=iso -- "$file")"
+    unix_timestamp="$(date --date="$full_time" +%Y%m%d%H%M.%S)"
+    touch -m -t "$unix_timestamp" "$file"
+done
+
+echo "===> Removing files that should not be uploaded to S3"
+# NOTE: This ensures that they will be deleted from the bucket if they're already there.
+# If we used `aws s3 sync --delete --exclude` instead, they would not get deleted.
+find -path './.*' -delete
+find -path './_*' -delete
+
+echo "===> Adding compatibility symlinks for files containing plus signs in the name"
+# NOTE: This is a quick'n'dirty workaround for Amazon S3 decoding plus sign in paths
+# as a space even though this substitution is only supposed to happen in a query string.
+# See https://forums.aws.amazon.com/thread.jspa?threadID=55746
+find . \
+    -regex "^\(.*/\)*[^/]*\+[^/]*$" \
+    -exec bash -c 'ln --symbolic --no-target-directory "$(basename "{}")" "$(dirname "{}")/$(basename "{}" | tr "+" " ")"' \;
+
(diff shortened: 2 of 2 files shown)


cameel:
Here's a bash script you can use to make sure that all files are actually available on S3 (you could actually use it against any host, it's not S3-specific). If you run it in your local checkout of `solc-bin` it will go through all the files, download the equivalents from S3 one by one and diff them.

For simplicity I did not make it use git ignore rules so make sure you have no stray unversioned files in there (e.g. `node_modules`). It does ignore files starting with dots and underscores though.

``` bash
#!/usr/bin/env bash

set -e

s3_url="https://solc-bin.s3.eu-central-1.amazonaws.com"
tmp_dir="$(mktemp --directory)"

readarray -t files < <(find . -type f,l -regex '\./[^._].*' -printf "%P\n" | sort)

missing_file_count=0
different_file_count=0
file_count=0

mkdir -p "$tmp_dir"
for file in "${files[@]}"; do
    printf "%s: " "$file"

    if ! curl "${s3_url}/${file}" --output "${tmp_dir}/${file}" --fail --silent --no-progress-meter --create-dirs; then
        echo MISSING
        ((++missing_file_count))
    else
        if ! cmp --silent -- "$file" "${tmp_dir}/${file}"; then
            echo DIFFERENT
            ((++different_file_count))
        else
            echo OK
        fi
    fi

    rm "${tmp_dir}/${file}"
    ((++file_count))
done
rm -r "$tmp_dir"

echo "Total files: ${file_count}"
echo "Missing:     ${missing_file_count}"
echo "Different:   ${different_file_count}"
```

cameel:
You can see the script in action in my fork of `solc-bin`: [cameel/solc-bin > actions > Mirror repository content to an S3 bucket](https://github.com/cameel/solc-bin/actions?query=workflow%3A"Mirror+repository+content+to+an+S3+bucket").

It's actually surprisingly fast. Even the full multi-GB upload takes only 5 min (plus another 5 for `solc-bin` checkout). I was expecting much worse.

I have the action in my fork hooked up to the `solc-bin` bucket right now so before merging this please remember to disable it - by removing current credentials of its user on AWS and generating new ones for this repo (as described above). Or just tell me to do it myself.

cameel:
Looks like the nightly action committing a new build does not trigger the S3 sync action. That's a limitation of github actions.

I see two possible workarounds:
1. Copy the S3 sync job to the nightly action. This adds some code duplication.
2. Schedule the S3 sync to run e.g. 1 hour after the nightly. No duplication but there's either unnecessary delay or we risk syncing too soon and missing the build if it takes longer than usual. It also requires us to remember to update the sync schedule if we ever change the nightly schedule.

chriseth:
I think scheduling at roughly the right time is the better solution. We do not depend on this having a small delay - even 24 hours would be acceptable short-term.

cameel:
Fine. Scheduled to run at 1:00 now.

And two more changes to account for S3 bucket updates not being atomic:
- Syncing `list.*` files last. Clients won't see updated list with new entries until all files are already in the bucket. At least as long as we only add new binaries and never remove old ones.
- Added `turnstyle` action to prevent running multiple sync operations at the same time (see https://github.com/ethereum/solc-bin/pull/47/files#r461004973).

.github/workflows/s3-mirror.yml
```diff
@@ -0,0 +1,41 @@
...
+  push-to-s3:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+
+    steps:
+      - name: Wait for other instances of this workflow to finish
+        # It's not safe to run two S3 sync operations concurrently with different files
+        uses: softprops/turnstyle@v1
+        with:
+          same-branch-only: no
```
cameel, line 27: Unfortunately there's no way to tell github to run only one instance of the action at a time. As a workaround I used the [turnstyle](https://github.com/softprops/turnstyle) action.

Is it OK to use it? I looked at its code briefly and I didn't see anything nefarious. The docs say that giving it `GITHUB_TOKEN` is required but seems to work without it just fine. So this PR should be completely safe (turnstyle can't access any secrets) but it's a third-party action so if we ever update it without reviewing the new code **and** we start putting any secrets (`GITHUB_TOKEN` or S3 keys) in env variables it could theoretically steal them and use them to modify files on github or in S3.

cameel:
Just one minor change: added `--no-progress` to `aws s3 sync` to get clearer logs.

chriseth (approved):

cameel:
Below is our current S3/cloudfront config.

I don't have a better place to put it and it's not important enough to be preserved in the repo but I'd like to be able to point someone at it if we ever need to revisit the configuration.

There's nothing here that could not be recreated with some trial and error but some values I set were different from defaults so having a record of it may make things easier for us in the future.

### S3 bucket
- Name: `solc-bin`

#### Bucket policy
``` json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "AddPerm",
            "Effect": "Allow",
            "Principal": "*",
            "Action": "s3:GetObject",
            "Resource": "arn:aws:s3:::solc-bin/*"
        },
        {
            "Sid": "AddPerm",
            "Effect": "Allow",
            "Principal": "*",
            "Action": "s3:ListBucket",
            "Resource": "arn:aws:s3:::solc-bin"
        }
    ]
}
```
EDIT 2021-01-07: originally the policy did not contain a permission for listing bucket contents and this made the server return HTTP 403 Forbidden (rather than 404) for missing files.
EDIT 2021-09-06: Amazon converted the config to JSON.

#### CORS configuration
```json
[
    {
        "AllowedHeaders": [
            "*"
        ],
        "AllowedMethods": [
            "GET",
            "HEAD"
        ],
        "AllowedOrigins": [
            "*"
        ],
        "ExposeHeaders": []
    }
]
```

### Cloudfront
Values in **bold** differ from defaults or values pre-filled by Amazon.

#### General
| Parameter                       | Value                                      |
|---------------------------------|--------------------------------------------|
| Domain Name                     | `d1ilmvg25bc920.cloudfront.net`            |
| Alternate Domain Names (CNAMEs) | **`solc-bin.ethereum.org`**                |
| SSL Certificate                 | **Custom SSL Certificate (example.com)**   |
| Default Root Object             | **`README.html`**                          |
| Supported HTTP Versions         | HTTP/2, HTTP/1.1, HTTP/1.0                 |
| IPv6                            | Enabled                                    |
| Delivery Method                 | Web                                        |
| Cookie Logging                  | Off                                        |
| Price Class                     | Use All Edge Locations (Best Performance)  |
| AWS WAF Web ACL                 | -                                          |
| Custom SSL Client Support       | -                                          |
| Security Policy                 | TLSv1.2_2021                               |

EDIT 2021-03-12: we're serving `README.html` rather than `README.md` now.
EDIT 2021-09-06: Security Policy changed from `TLSv1` to `TLSv1.2_2021` (which might be the default now)

#### Origins and Origin Groups
| Parameter                       | Value                                      |
|---------------------------------|--------------------------------------------|
| Origin ID                       | `S3-solc-bin`                              |
| Origin Domain Name              | **`solc-bin.s3.amazonaws.com`**            |
| Origin Path                     |                                            |
| Restrict bucket access          | No                                         |

| Parameter                       | Value                                      |
|---------------------------------|--------------------------------------------|
| Origin ID                       | `S3-solc-bin-with-static-Origin-header`    |
| Origin Domain Name              | **`solc-bin.s3.amazonaws.com`**            |
| Origin Path                     |                                            |
| Restrict bucket access          | No                                         |
| Add custom header               | **`Origin`: `https://static-origin.example.com`** |

EDIT 2021-09-14: Added `S3-solc-bin-with-static-Origin-header` origin.

#### Behaviors
- Defaults for all paths
    | Parameter                         | Value                                      |
    |-----------------------------------|--------------------------------------------|
    | Path Pattern                      | Default (`*`)                              |
    | Cache Policy                      | **`Managed-CachingOptimized`**             |

All behaviors have these settings in common:
| Parameter                         | Value                                          |
|-----------------------------------|------------------------------------------------|
| Origin or Origin Group            | **`S3-solc-bin-with-static-Origin-header`**    |
| Viewer Protocol Policy            | HTTP and HTTPS                                 |
| Allowed HTTP Methods              | **GET, HEAD, OPTIONS**                         |
| Field-level Encryption Config     | -                                              |
| Cached HTTP Methods               | GET, HEAD, **OPTIONS**                         |
| Cache and origin request settings | Use a cache policy and origin request policy   |
| Origin Request Policy             | **`Managed-CORS-S3Origin`**                    |
| Smooth Streaming                  | No                                             |
| Restrict Viewer Access            | No                                             |
| Compress Objects Automatically    | **Yes**                                        |

EDIT 2021-09-06: Removed the behavior that disabled caching for list files.
EDIT 2021-09-14: Switched origin to `S3-solc-bin-with-static-Origin-header` and enabled caching for `OPTIONS` requests.

PR #49, opened 2020-07-27
Creating wasm/ symlinks in the update script

cameel:
Depends on #48. Part of https://github.com/ethereum/solidity/issues/9258.

It used to be the [responsibility of `publish_binary.sh`](https://github.com/ethereum/solidity/blob/v0.6.12/scripts/travis-emscripten/publish_binary.sh#L90) but this script is not a part of the new, manual process of adding binaries to `solc-bin`. I made `update` do it instead.

### DIFF
--- a/update
+++ b/update
@@ -207,7 +207,6 @@ function processDir (dir, listCallback) {
 
 const DIRS = [
   '/bin',
-  '/emscripten-asmjs',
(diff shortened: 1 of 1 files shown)


chriseth:
Rebased.

chriseth (approved):

PR #50, opened 2020-07-28
Add --reuse-hashes option to update script

cameel:
Part of https://github.com/ethereum/solidity/issues/9258. The PR is based on #49 but only for convenience - both modify `processDir()` function and would conflict otherwise.

You can now run `./update --reuse-hashes` and have the script finish almost immediately if `list.json` files already exist.

Note that the `update` script usually truncates the lists just after it starts and only then all the heavy processing happens. That's probably due to lazy evaluation of `parsedList`. This means that if you interrupt it or if it fails for some reason, you should revert the lists using `git` or it will have to recalculate all hashes anyway.

### DIFF
--- a/update
+++ b/update
@@ -74,10 +74,6 @@ function updateCopy (srcRelativeToRoot, destRelativeToRoot) {
   })
 }
 
-function readFile (dir, file) {
-  return fs.readFileSync(path.join(__dirname, dir, file))
-}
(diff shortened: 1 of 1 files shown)


chriseth:
Rebased.

chriseth (approved):

PR #53, opened 2020-09-28
Github action for building solc 0.3.6 - 0.6.8 on MacOS

cameel:

This PR adds a github action that randomly selects one of the missing MacOS binaries, builds it and pushes it to a branch. When merged it will start building on schedule (every 15 minutes) until they're all available. The branch is meant to be later submitted in a separate PR (manually, after updating file lists). Then the action can be disabled or removed.

I also have the binaries already built in my own fork (#54) so maybe we want to take them instead.
I'm submitting it anyway so that I can get it reviewed and to preserve the script in case we ever need to rebuild again.

Points of note:
- I was able to make it work for versions down to 0.3.6. Below that it should still be doable but it's much more complex. The build configuration is shared with old versions of `cpp-ethereum` and there are many dependencies that are hard to track down in binary form and would have to be built from source.
- Below 0.5.13 there's no script for easily installing dependencies and I'm always linking with Z3 4.8.5.
- I managed to statically link even versions below 0.4.6 by backporting a change that added dynamic linking for jsoncpp.
    - Note that MacOS builds are not completely static. They all still dynamically link to `libSystem` and `libc++` but that's the case even with 0.7.x builds.
- The script runs tests for each version but in the older ones the tests required a running cpp-ethereum/aleth node. I was not able to run these. Versions that are even older only have command-line tests or no tests at all - in that case I'm just doing a simple smoke test of running the compiler on a minimal contract. SMT tests are also disabled in old versions (not sure about newer ones).
- I did not do bytecode comparisons.

### DIFF
--- a/.github/workflows/random-macosx-build.yml
+++ b/.github/workflows/random-macosx-build.yml
@@ -1,0 +1,322 @@
+name: MacOS build of a random missing solc version
+
+on:
+  push:
+    # Action disabled. The filter excludes all the possible branch names.
+    branches: ["!*"]
+
+env:
+  MIN_SUPPORTED_SOLIDITY_VERSION: 0.3.6
+  TARGET_BRANCH: macosx-static-binaries
+  COMMITTER_NAME: macosx build action
+  COMMITTER_EMAIL: builds@ethereum.org
+
+jobs:
+  select-solc-version:
+    runs-on: ubuntu-latest
+    outputs:
+      solidity-version: ${{ env.SOLIDITY_VERSION }}
+
+    steps:
+      - name: Install semver utility
+        run: |
+          npm install semver
+
+      - name: Select version to build
+        run: |
+          github_tag_response=$(curl "https://api.github.com/repos/ethereum/solidity/git/refs/tags/")
+          github_macosx_files_response=$(curl "https://api.github.com/repos/${GITHUB_REPOSITORY}/contents/macosx-amd64?ref=${TARGET_BRANCH}")
+
+          available_tags=$(
+            echo "$github_tag_response" |
+            jq --raw-output '.[].ref' |
+            sed -n 's|^refs/tags/v\([0-9.]\+\)$|\1|p'
+          )
+          eligible_tags=$(
+            npx semver --range ">= ${MIN_SUPPORTED_SOLIDITY_VERSION}" $available_tags |
+            sort
+          )
+          available_binaries=$(
+            echo "$github_macosx_files_response" |
+            jq --raw-output '.[].name' |
+            sed -n 's/^solc-macosx-amd64-v\([0-9.]*\)+commit.[0-9a-f]\+$/\1/p' |
+            sort
+          )
+          missing_versions="$(comm -23 <(echo "$eligible_tags") <(echo "$available_binaries"))"
+          selected_version=$(echo "$missing_versions" | sort --random-sort | head --lines 1)
+
+          echo "Missing Solidity versions:"
+          echo "${missing_versions}"
+          echo "Selected version: ${selected_version}"
+
+          echo "SOLIDITY_VERSION=${selected_version}" >> $GITHUB_ENV
+
+  build:
+    runs-on: macos-10.15
+    needs:
+      - select-solc-version
+    env:
+      SOLIDITY_VERSION: ${{ needs.select-solc-version.outputs.solidity-version }}
+    outputs:
+      full-build-version: ${{ env.FULL_BUILD_VERSION }}
+
+    if: "needs.select-solc-version.outputs.solidity-version"
+    steps:
+      - name: Check out Solidity source
+        uses: actions/checkout@v2
+        with:
+          repository: 'ethereum/solidity'
+          ref: v${{ env.SOLIDITY_VERSION }}
+          path: solidity/
+          fetch-depth: 0
+
+      - name: Install semver utility
+        run: |
+          npm install --global semver
+
+      - name: Install dependencies
+        run: |
+          cd solidity/
+
+          if semver --range '>= 0.5.13' "$SOLIDITY_VERSION"; then
+            test -f .circleci/osx_install_dependencies.sh
+
+            # On 0.5.13 and higher we can just use the script for installing dependencies for CI
+            .circleci/osx_install_dependencies.sh
+          else
+            test ! -e .circleci/osx_install_dependencies.sh
+
+            brew unlink python
+            brew install coreutils
+
+            # Not all boost versions work with all solc versions. Through trial and error I found
+            # out that boost 1.65.0 works correctly with every solc version since 0.4.0 and possibly
+            # with earlier ones too.
+            # NOTE: Currently homebrew provides only boost 1.73.0 and the easiest way to build 1.65.0 is to
+            # check out and older formula directly from their repo.
+            curl --remote-name https://raw.githubusercontent.com/Homebrew/homebrew-core/710e281d25b5f9b583f0e51ef33cca797c27c909/Formula/boost.rb
+            brew install ./boost.rb
+
+            # NOTE: As long as the build does not fail, the compiler should work with any Z3 version.
+            # The prover might find slightly different answers but they should not be wrong.
+            # SMT tests are tied to a specific Z3 version and might fail due to that but here
+            # we're only running them when building versions later than 0.5.13 so it's not an issue.
+            z3_version="4.8.5"
+            z3_release="z3-${z3_version}-x64-osx-10.14.2"
+            curl -O -L "https://github.com/Z3Prover/z3/releases/download/Z3-${z3_version}/${z3_release}.zip"
+            unzip "${z3_release}.zip"
+            rm "${z3_release}.zip"
+            mv "${z3_release}/bin/libz3.a" /usr/local/lib
+            mv "${z3_release}/bin/z3"      /usr/local/bin
+            mv "${z3_release}/include/"*   /usr/local/include
+            rm -r "${z3_release}"
+
+            # Since 0.5.11 evmone can be used as an altenative to IPC tests
+            if semver --range '>= 0.5.11' "$SOLIDITY_VERSION"; then
+              test -f test/EVMHost.h
+
+              evmone_version="0.1.0"
+              evmone_package="evmone-0.1.0-darwin-x86_64"
+              curl -O -L "https://github.com/ethereum/evmone/releases/download/v${evmone_version}/${evmone_package}.tar.gz"
+              tar xzpf "${evmone_package}.tar.gz" -C /usr/local
+              rm "${evmone_package}.tar.gz"
+            else
+              test ! -e test/EVMHost.h
+            fi
+          fi
+
+      - name: Determine full build version
+        run: |
+          cd solidity/
+
+          # Can't run get_version.sh because MacOS grep does not have the -P flag.
+          test "$SOLIDITY_VERSION" = "$(sed -n 's/^set(PROJECT_VERSION "\([0-9.]*\)")$/\1/p' CMakeLists.txt)"
+
+          last_commit_hash=$(git rev-parse --short=8 HEAD)
+          full_build_version="v${SOLIDITY_VERSION}+commit.${last_commit_hash}"
+
+          echo "FULL_BUILD_VERSION=${full_build_version}" >> $GITHUB_ENV
+
+      - name: Build
+        run: |
+          cd solidity/
+
+          # In 0.3.6 boostTest.cpp has its own main() function which leads to "duplicate symbol '_main'" linker error
+          # despite BOOST_TEST_NO_MAIN being defined. 0.4.0 does not have this problem so here we just backport that change.
+          if semver --range '= 0.3.6' "$SOLIDITY_VERSION"; then
+            # NOTE: Only the first commit should be necessary but one of its changes seems to
+            # have been mistakenly included in the other and it won't compile on its own.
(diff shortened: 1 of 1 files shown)


cameel:
Here's a dump of my notes from working on these builds. Might help someone trying to build an older version on his Mac.

### Dependencies
#### Static linking
Builds on OSX are not completely static. These two dynamic libraries are always present:
```
/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 902.1.0)
/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
```

#### Z3
Our CMake config can link Z3 both dynamically and statically. If both are available, it depends on which is returned first by CMake's `find_library()`. In my builds it seems that the dynamic one gets chosen by default.
- Z3 from homebew provides both the dynamic and the static library.
- `osx_install_dependencies.sh` forces static linking by downloading a binary release of Z3 from github and only installing the static library.
- Similarly for Ubuntu we have a PPA that has a package containing only static Z3.

solc links against Z3 since vesion 0.4.16 where "experimental partial support for z3 smt checker" was introduced.

#### Boost
- Homebrew currently provides only the latest version of boost (1.73.0), which won't work with older solc (see https://github.com/ethereum/solidity/issues/6349#issuecomment-694823895).
- There are formulae for boost 1.60.0 and older but they're disabled (`Error: boost@1.60 has been disabled because it is a versioned formula!`).
- It's easy to build earlier versions by downloading earlier versions of the formula and running them locally
    - [Boost 1.70.0 formula](https://raw.githubusercontent.com/Homebrew/homebrew-core/5da5895add2f6b9320d654dd844d4827f6876c8b/Formula/boost.rb) is still too new (see https://github.com/ethereum/solidity/issues/6349#issuecomment-694990023)
    - [Boost 1.60.0 formula](https://raw.githubusercontent.com/Homebrew/homebrew-core/e0ecc9519ca9c0491a3ec336985b93cf1705c9b4/Formula/boost.rb) fails because it depends on something else in homebrew that has been moved/removed.
    - [Boost 1.63.0 formula](https://raw.githubusercontent.com/Homebrew/homebrew-core/be7a0f0ac45c03e453c7940df79ce2f608d0ebaf/Formula/boost.rb) worked for most of my builds but I eventually switched to 1.65.0.
    - [Boost 1.65.0 formula](https://raw.githubusercontent.com/Homebrew/homebrew-core/710e281d25b5f9b583f0e51ef33cca797c27c909/Formula/boost.rb) works with all solc versions between 0.3.6 and 0.6.8. Likely with newer ones too.

#### jsoncpp
CMake config will download and build jsoncpp on its own. This works at least since 0.3.6, probably earlier. If you install the latest one with `brew` in addition to that (currently 1.9.3), your version will be used instead and, below 0.4.22, the build will fail with `Undefined symbols for architecture x86_64` linker error. 0.4.23 will just refuse to use anything other than jsoncpp 1.7.7.

Version 0.5.0 adds `scripts/install_obsolete_jsoncpp_1_7_4.sh`, which automates the installation of older jsoncpp (https://github.com/ethereum/solidity/commit/54f83acee971a72028dda5d79ba3d07c219eb2fd). This version of jsoncpp works fine too (except for solc 0.4.23) but installing it doesn't seem really necessary. BTW, `./circleci/osx_install_dependencies.sh` is used on CircleCI and runs this script it since 0.5.13.

### Version-specific

#### solc 0.6.1
solc 0.6.1 and later compile without any workarounds.

#### solc 0.6.0
- This version stopped relying on buggy behavior of boost < 1.71.0 (See https://github.com/boostorg/multiprecision/issues/129) regarding serialization of `multiprecision` numbers as hexadecimal numbers. On solc versions earlier than this tests won't pass when compiled with boost >= 1.71.0. See https://github.com/ethereum/solidity/issues/6349#issuecomment-694823895 for details.

#### solc 0.5.13
- `.circleci/osx_install_dependencies.sh` introduced (https://github.com/ethereum/solidity/commit/160983141fa66e2cf112594638bd81416a2a9088).
    Before that `.circleci/config.yml` was installing latest `z3` from homebrew and `evmone` was not being installed at all.
- Before this version, `scripts/tests.sh` was hard-coded to run with the `--no-smt` flag on MacOS (https://github.com/ethereum/solidity/commit/c9a56d5339a872cd90014cb686ca0b56eac499cf).

#### solc 0.5.12
- There are tons of linker warnings like this in the log, but nothing appears to be actually broken:
    ```
    ld: warning: direct access in function (..) means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
    ```

#### solc 0.5.11
- The first version to use `evmone` for testing.
- `--no-ipc` flag no longer available and not needed in `soltest`.
- `direct access` linker warnings like in 0.5.12.

#### solc 0.5.10
- `scripts/tests.sh` no longer runs `soltest` with `--no-ipc` flag on MacOS.
    - Due to this it tries to install and run a Linux version of aleth, then hangs. A workaround is to change hard-coded `IPC_ENABLED` value to `false` and `IPC_FLAGS` to `--no-ipc`.
    - If `realpath` is not installed (included in the `coreutils` homebrew package), it also prints an error about that (but still continues).
- `direct access` linker warnings like in 0.5.12.

#### solc 0.5.8 - 0.5.9
- `direct access` linker warnings like in 0.5.12.

#### solc 0.5.0
- This version switched from cpp-ethereum to [aleth](https://github.com/ethereum/aleth) for IPC tests in `scripts/tests.sh` (https://github.com/ethereum/solidity/commit/273be204b6b2e96a5dfe8f260b913987ab4e3183).
- This version fixes a bug in the way boost unit tests headers are included. On 0.4.x this bug causes linker errors (`duplicate symbol`). The workaround is to replace `#include <boost/test/included/unit_test.hpp>` with `#include <boost/test/unit_test.hpp>` in `test/boostTest.cpp`.
    - Reported in https://github.com/ethereum/solidity/issues/3845. Fixed in https://github.com/ethereum/solidity/issues/4572.
    - Docs on the usage of these headers: https://www.boost.org/doc/libs/1_74_0/libs/test/doc/html/boost_test/usage_variants.html
    - In https://github.com/ethereum/solidity/issues/3845 someone claims that just switching from boost 1.66.0 to boost 1.65.1 helped. For me the errors appeared on both boost 1.63.0 and boost 1.65.0.
- Commandline tests are broken on MacOS due to buggy sed call in `test/cmdlineTests.sh`.
    ```
    too_long_line.sol:2:164: Error: Identifier not found or not unique.
    Commandline tests FAILED
    ```
    This was fixed in 0.5.1 (https://github.com/ethereum/solidity/pull/5419) and the fix can be cleanly backported.

#### solc 0.4.24
- Starting with this version `scripts/tests.sh` runs `soltest` with `--no-ipc` flag on MacOS (https://github.com/ethereum/solidity/commit/743a714f21c76c0c2a026b4ceb61c7cf94130540)
- This version and later ones build without errors with jsoncpp 1.9.3.

#### solc 0.4.23
- This particular version will build only with jsoncpp 1.7.7. It has a `debug_assert()` that fails otherwise.
    ```
    /Users/runner/work/solc-bin/solc-bin/solidity/libdevcore/JSON.cpp:30:1: error: static_assert failed due to requirement '(1 == 1) && (9 == 7) && (3 == 7)' "Unexpected jsoncpp version: 1.9.3. Expecting 1.7.7."
    ```

#### solc 0.4.22
- Tests in CI now run also against `constantinople` EVM (https://github.com/ethereum/solidity/commit/44500341748070bd1ad97046b915bf2c76b1eb9b).

#### solc 0.4.21
- Ability to select between `byzantium` and `homestead` EVM introduced (https://github.com/ethereum/solidity/commit/83515eadcf21decc9355c69c621c4d530a62b04e).

#### solc 0.4.17
- `deps/` git submodule removed.

#### solc 0.4.16
- `STATIC_LINKING` option in CMake replaced with `SOLC_LINK_STATIC` which has the same function (https://github.com/ethereum/solidity/commit/4974ff3962759a19150877b0e59e64319190ba1d).

#### solc 0.4.12
- `scripts/get_version.sh` added (https://github.com/ethereum/solidity/commit/f13feed7f26aba22396ec1728a51a96fd9bd65e3).
    - It's not usable on MacOS anyway because it uses `-P`/`--perl-regexp` option with `grep`, which is not available on that system.

#### solc 0.4.11
- Introduces `--no-ipc` option in soltest (https://github.com/ethereum/solidity/commit/ea6d925a5da79288bc2a3197c30f11c895a07a1a).

#### solc 0.4.10
- `scripts/tests.sh` runs `solfuzzer` (https://github.com/ethereum/solidity/commit/8be318e75bbed68520d3a3ef34f777804b6b60e0).

#### solc 0.4.9
- Introduces `cmdlineTests.sh` (https://github.com/ethereum/solidity/commit/605455f96b64f0c4b229cc31186f2dcf22433e6d). `scripts/tests.sh` runs it.

#### solc 0.4.5
- `deps/` submodule updated to a revision that contains `jsoncpp.cmake` module which downloads and builds jsoncpp from source at cmake configuration stage (https://github.com/ethereum/solidity/pull/1252).

#### solc 0.4.2 - 0.4.4
- These versions download and install prebuilt jsoncpp from https://github.com/ethereum/cpp-dependencies/releases/download/vc140/jsoncpp-1.6.2.tar.gz.
- jsoncpp is linked dynamically. https://github.com/ethereum/solidity/pull/1252 can be cleanly backported to get static linking with 1.7.7 instead.

#### solc 0.4.0 - 0.4.1
- Same situation with jsoncpp as in 0.4.2 but https://github.com/ethereum/solidity/pull/1085 needs to be backported too due to conflicts.

#### solc 0.3.6
- `deps/` git submodule introduced.
- `tests/boostTest.cpp` has a `main()` function. Due to this linking fails with `duplicate symbol '_main'`, even though it defines `BOOST_TEST_NO_MAIN`.
    - This affects only 0.3.6. Neither 0.3.5 nor 0.4.0 has its own `main()`.
    - https://github.com/ethereum/solidity/pull/837 from 0.4.0 removes `main()` and can be cleanly backported to 0.3.6.
- Same situation with jsoncpp as in 0.4.0 but there are a few more PRs that needs to be backported due to conflicts: https://github.com/ethereum/solidity/pull/1035, https://github.com/ethereum/solidity/pull/975, https://github.com/ethereum/solidity/pull/914.

#### solc 0.3.0
- CMake configuration depends on https://github.com/ethereum/webthree-helpers. The necessary files have been removed from the latest revision, though they're still available in history.
- At this point cmake configuration was shared with `cpp-ethereum`. There are lots of extra dependencies and most of them probably not really needed for Solidity.

#### solc 0.1.2
- CMake configuration depends on https://github.com/ethereum/cpp-ethereum-cmake

### Attempts to run IPC tests with cpp-ethereum/aleth
- Tests in 0.4.0 fail with [Aleth 1.5.0 binary release for MacOS](https://github.com/ethereum/aleth/releases/download/v1.5.0/aleth-1.5.0-darwin-x86_64.tar.gz)
    ```
    /Users/runner/work/solc-bin/solc-bin/solidity/test/RPCSession.cpp:312: fatal error: in "SolidityAuctionRegistrar/creation": Error on JSON-RPC call: INTERNAL_ERROR: : /Users/distiller/project/libethereum/ClientTest.cpp(72): Throw in function void dev::eth::ClientTest::setChainParams(const string &)
    Dynamic exception type: boost::exception_detail::clone_impl<dev::eth::ChainParamsInvalid>
    [dev::tag_comment*] = /Users/distiller/project/libdevcore/JsonUtils.cpp(89): Throw in function void dev::requireJsonFields(const json_spirit::mObject &, const std::string &, const std::map<std::string, JsonFieldOptions> &)
    Dynamic exception type: boost::exception_detail::clone_impl<dev::MissingField>
    [dev::tag_comment*] = Expected field 'difficulty' not found in config: ChainParams::loadConfig::genesis
    ```
    - That's because in that version of aleth the `difficulty` field in the RPC message that creates the genesis block became mandatory. Solidity started adding this field in 0.5.0 (https://github.com/ethereum/solidity/commit/0319b184ce2e297584097e92a16cb695255f4246). IPC tests in earlier solc versions will only run with earlier versions of aleth.
- [Aleth 1.2.9 binary release for MacOS](https://github.com/ethereum/webthree-umbrella/releases/tag/v1.2.9) won't run because it's linked against system-wide `libsnappy.1.dylib` rather than the one included in the package.
    ```
    dyld: Library not loaded: /usr/local/lib/libsnappy.1.dylib
    Referenced from: /Users/runner/work/solc-bin/solc-bin/cpp-ethereum/ethbin/libleveldb.1.dylib
    Reason: image not found
    ```
- [Aleth 1.4.0](https://github.com/ethereum/aleth/releases/tag/v1.4.0) is not statically linked and the release package does not include the dependencies.
- Aleth 1.3.0 is referenced in various places, including Reddit ([RELEASED - cpp-ethereum-v1.3.0 ("Homecoming")](https://www.reddit.com/r/ethereum/comments/4tk44x/released_cppethereumv130_homecoming/)) but the binary release is gone.    It's not in the `aleth` repo and the `cpp-ethereum` repo is now just a redirect to it.

chriseth:
Looks good! I'll merge the PR containing the binaries. Please deactivate this job, then it can be merged.

cameel:
@chriseth Sure. I'll do that tomorrow. Do we want to keep it after that or just merge to have it in history and then remove?

chriseth:
Let's keep it - if it's just in the history, nobody will remember it is there.

cameel:
Done and rebased. Now the action won't ever run unless someone modifies it.

cameel:
Action updated not to use `set-env` (see #60).

chriseth (approved):

PR #58, opened 2020-10-05
Async file read to lower memory usage in update script

cameel:
Workaround for https://github.com/ethereum/solidity/issues/9956.

This is basically a hack. Or at most a harmless optimization that we should not really rely on.

Currently IPFS hash computation is queued for asynchronous execution only after the whole file has been read into memory and processed using keccak and swarm's hashing function. This results in all the read operations being executed first and hogging a lot of memory that can't be freed until IPFS hashing finishes. Making read asynchronous allows the IPFS hashing to be queued immediately and be interleaved with read operations. This lowers peak memory usage to about 6 GB on my machine. I think this should be enough to at least stop crashing the nightly builds even if it doesn't solve the problem completely.


ekpyron (approved):
Let's just give it a try.

PR #62, opened 2020-10-07
Add missing wasm files.

chriseth:


cameel:
Is it a bug in the `update` script or did you just forget to commit them? Looks like the latter because the files were already on the lists in #61 but asking just in case.

chriseth:
No, I forgot to add them to the manual commit.

PR #66, opened 2020-11-18
Move PR checks to CircleCI

### LINKED ISSUE #9982: Move solc-bin PR checks from Travis to Gihub Actions or CircleCI
Related to #9931 and #9932.

Just executing the `update` script takes 18 minutes on Travis. In nightly builds on GH actions it's just 13 minutes. Might be even faster in CircleCI but setup might require more work.

cameel:
Resolves https://github.com/ethereum/solidity/issues/9982

Still a draft. There will be some random activity in this PR for a moment here because I need to check if it works.


cameel:
`chk_npm_test` is failing and I'm afraid it's the same problem as in https://github.com/ethereum/solidity/issues/9956.

```
> solc-bin@0.0.0 update
> ./update

Processing directory /bin
Processing directory /linux-amd64
Processing directory /macosx-amd64
Processing directory /windows-amd64
Computing hashes of '/linux-amd64/solc-linux-amd64-v0.4.10+commit.f0d539ae'
Computing hashes of '/linux-amd64/solc-linux-amd64-v0.4.11+commit.68ef5810'
Killed
npm ERR! code 137
npm ERR! path /home/circleci/project
npm ERR! command failed
npm ERR! command sh -c ./update

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/circleci/.npm/_logs/2020-11-18T12_12_27_030Z-debug.log
npm ERR! code 137
npm ERR! path /home/circleci/project
npm ERR! command failed
npm ERR! command sh -c git checkout -f && npm run update && git diff --exit-code

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/circleci/.npm/_logs/2020-11-18T12_12_27_055Z-debug.log


Exited with code exit status 137

CircleCI received exit code 137
```

cameel:
I followed the suggestion to use a bigger machine on CircleCI and `xlarge` is enough to run these PR checks. In that case I'm closing #67 in favor of this PR and I'm marking the PR as ready for review.

The checks even passed on `large` initially but I ran them several times and 1 in 3 failed. On `medium` even a single run fails.

cameel:
Also, they are indeed much faster than on Github actions. 2 min for checkout + 5:30 min for the `update` script.

chriseth:
Can we make use of the 16 (?) cores the xlarge container has?

cameel:
It's only 8 cores. `2xlarge` has 16. I think we're using them all because the `update` script runs hundreds or parallel hashing jobs simultaneously. The ~3.5x speed improvement over Github Actions (which only has 2 cores) is in the right ballpark. That's probably why it finishes in 5 minutes rather than 18.

Though something still seems a bit off because I think the running time was also 5 minutes in those runs of the `large` instance that did not crash. Is it worth investigating?

chriseth:
Why do you think it is running on parallel? If you just use regular nodejs, everythingis sequential, as far as I know.

chriseth:
I just ran it locally, and the nodejs process used 13gigs of virtual memory and 5.6 resident - but maybe this is just the garbage collector. I'm wondering if invoking a real binary for the hashing would be quicker...

chriseth:
If we are lucky, only one of the three hashes is slow.

chriseth:
Another thing to consider: If we use async IO, then it could be that all files are read into RAM at the same time before the first hash is actually computed...

cameel:
Ah, you got me :) I forgot that the async jobs all run in the same thread.

We could make it parallel though. The most straightforward way to use more cores would be to make it possible to run the script on each directory separately and then to run several instances in parallel. Alternatively, we could make the script use real threads internally but I think that would be harder. At least for me because I have never used threads in JS so I don't know the libraries.

But without it, yeah, using a `large` instance is a waste. It was just meant as a quick and dirty workaround for https://github.com/ethereum/solidity/issues/9956 anyway because I really don't want to get into that issue right now. I just wanted to move the checks off Travis and make them a bit faster. If we want to polish it I'd create an issue about it for later because I'm currently getting too deep into these small nice-to-have tasks that keep piling up. I'd like to unwind my stack a bit and finish some older ones :) If this instance is too large, I'd rather just put it on Github Actions for now.

cameel:
>  If we use async IO, then it could be that all files are read into RAM at the same time before the first hash is actually computed...

They are. Not all but most of them. That's what https://github.com/ethereum/solidity/issues/9956 is all about - how many files get loaded into memory depends only on the ordering of the async jobs and there's nothing preventing it from just trying to load them all first and only then doing the hashing.

We should be using some kind of job pool that lets only a limited number run simultaneously instead.

.circleci/config.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - checkout
+      - run: npm install
+      - run: npm run test
+
+workflows:
+  version: 2
+  workflow:
+    jobs:
+      - chk_npm_lint
```
chriseth: It takes 2 minutes to do the code checkout. Wouldn't it be better to run `npm run lint` just as part of the other job?

cameel: It would be possible, not sure if better. Both have pros and cons. I think it's more convenient to get simultaneous feedback about both linter issues and test failures instead of having to rerun it after fixing one. It also does not affect the overall time because they run in parallel.

But if you mean that it's a bit wasteful - yeah, combining them would definitely use less resources, especially now that I changed it to a `xlarge` machine (though lint still runs in `medium` and might even fit in the smallest one). I thought that it does not matter all that much since there is not much PR traffic in this repo anyway. Should I combine them?

chriseth: The thing is: All CI minutes do cost money :)

chriseth: On the other hand, I think this repo only gets an update once a day, so it might not be that bad after all.

cameel: I know, but if minutes were the only thing we care about, we'd just move it to Github Actions where we don't pay for them :) And having separate feedback from test and linter means that you can fix both without a rerun in the middle which could save some minutes.

But I'm not really dead set on having them as separate checks. I had a hard time deciding between both options in the first place and I just thought this one is *slightly* nicer. I have just pushed an update that merges the jobs.

So anyway, it would be nice to move off Travis before the next release. Which option do you prefer:
1. Merge this and be done with it but pay for an underutilized `xlarge`.
2. Resurrect #67 and go back to Github Actions.
3. Actually spend some time to optimize the `update` script right now (either by reducing its memory footprint so that we can use a smaller machine or parallelizing it so that `xlarge` gets fully utilized). We'll likely want to do that anyway so it does make sense, I just really wanted to focus on other tasks right now.
4. Stay on Travis until we optimize `update`.

My recommendation would be to go with option 2 but also keep this PR open and switch to CircleCI once we have the script optimized. I'm also fine with staying on Github Actions permanently if you think that's good enough.

chriseth (approved):

PR #67, opened 2020-11-18
Move PR checks to Github Actions

cameel:
Part of https://github.com/ethereum/solidity/issues/9982 but does not solve it because CircleCI would be the preferred solution in my opinion. #66 didn't work due to memory usage so trying Github Actions now.


cameel:
Bad news: the check took 18 minutes on first run so as much as on Travis.

Re-enabling Travis to compare and see if that one might be taking longer too due to some external factors but for now this looks bad.

cameel:
Travis fails due to a tiny bug in `swarmhash`: https://github.com/axic/swarmhash/pull/4.

The bug can only be triggered on the JS implementation of SHA-3 (`browserify-sha3`) so the C-based `sha3` must be failing to build. I'm not sure why now. There's nothing about it in the output.

cameel:
So anyway, due to the bug I can't compare the time with Travis but it does run much slower after all. Github Action-based checks consistently take 17-18 minutes. Travis claims that it ran for 11 minutes but I was checking the job and did not finish for at least 15 minutes after the GA jobs so this must be incorrect. It did not finish for at least 30 minutes for whatever reason. And it didn't even run the `update` script to completion (it failed on the first hashing operation).

Here are links to two failed Travis builds in case anyone wants to take a lot because I'm removing the job now (I want to mark the PR as ready for review):
- [build #215](https://travis-ci.org/github/ethereum/solc-bin/builds/744413140)
- [build #218](https://travis-ci.org/github/ethereum/solc-bin/builds/744429927)

cameel:
> Bad news: the check took 18 minutes on first run so as much as on Travis.

Actually, it **is** faster than Travis. In https://github.com/ethereum/solidity/issues/9982 I wrote that the nightly action takes 13 minutes but that's just to run the `update` script. With the checkout of solc-bin it's 17-18 minutes. I just checked nightly action runs to verify and yeah, it takes the same time there as in the PR checks here. I've seen `update` take anywhere from 10 to 14 minutes in nightly runs.

cameel:
I'm closing this in favor of #66 because I managed to get it to run on CircleCI by using a bigger machine.

PR #70, opened 2020-12-16
Make rebuilt macOS binaries executable

cameel:
I noticed that my rebuild script (#53) does not mark the binaries as executable before committing them and now I constantly have to do it manually when working with them. It would be best to have that bit set in the repo.


.github/workflows/random-macosx-build.yml
```diff
@@ -316,6 +316,7 @@ jobs:
       - name: Commit and push the binary
         run: |
           cd solc-bin/
+          chmod 0755 ../solc
```
axic, line 319: So basically the `download-artifact` step doesn't apply this (it doesn't convey this information).

axic (approved):

.github/workflows/random-macosx-build.yml
cameel: Yeah. Artifacts seem to be stored as zips internally and zips do not preserve the flags so it might be due to that.

Another possibility is that it could be just our build system not making it executable. But I think it's unlikely and I didn't want to investigate that too deeply. Putting `chmod` here will work in either case.

PR #73, opened 2020-12-22
macOS build action workarounds for newer homebrew and clang

cameel:
Just a bunch of workarounds to be able to still use the action.

The old way of installing boost no longer works because homebrew has removed the option to install from a local script. You can now kinda do this by creating a tap and you can kinda create a local one (even though Homebrew pushes you hard towards publishing it on github) but with so many workarounds needed, it was actually easier to just build from source.

Also, unfortunately newer Clang also requires patching Boost build system to be able to build it.

### DIFF
--- a/.github/workflows/random-macosx-build.yml
+++ b/.github/workflows/random-macosx-build.yml
@@ -90,12 +90,44 @@ jobs:
             brew install coreutils
 
             # Not all boost versions work with all solc versions. Through trial and error I found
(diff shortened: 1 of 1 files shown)


cameel:
Closing in favor of #74 - same thing but submitted as a PR directly from `solc-bin` rather than from my fork. There are weird problems with basing branches on one another when submitting from a fork.

PR #76, opened 2020-12-22
macOS binaries with correct version strings

### LINKED ISSUE #10183: Solidity 0.5.17 installed with Homebrew produces a different output than the version downloaded from solc-bin.ethereum.org
Hi

@udwig, reported to us that he was unable to verify a contract on Etherscan, even when using the exact same solidity standard input json. We spent some time debugging it, and found this problem with solc or with its distribution:

If you install `solc` `0.5.17+commit.d19bba13` using Homebrew and downloading it from solc-bin.ethereum.org, you can get a different output for the same input.

Here's a script that can be used to reproduce it:

```bash
brew tap ethereum/ethereum
brew install ethereum/ethereum/solidity@5

wget --quiet https://gist.githubusercontent.com/alcuadrado/06c1ad0cf7e23758e17a32e829f26497/raw/b9f864cfda21d0cc9c5365fa1a1b435aec6285e8/input-to-get-different-outputs.json

wget --quiet https://solc-bin.ethereum.org/macosx-amd64/solc-macosx-amd64-v0.5.17+commit.d19bba13
chmod +x solc-macosx-amd64-v0.5.17+commit.d19bba13

echo '\n\n\n'
echo Brew solc
solc --version
cat input-to-get-different-outputs.json | solc --standard-json | md5sum

echo '\n\n\n'
echo Native solc
./solc-macosx-amd64-v0.5.17+commit.d19bba13 --version
cat input-to-get-different-outputs.json | ./solc-macosx-amd64-v0.5.17+commit.d19bba13 --standard-json | md5sum
```

It will output

```
Brew solc
solc, the solidity compiler commandline interface
Version: 0.5.17+commit.d19bba13.Darwin.appleclang
b2051480575328ea24cdfa7cbdc71cf5  -




Native solc
solc, the solidity compiler commandline interface
Version: 0.5.17+commit.d19bba13.mod.Darwin.appleclang
a13020de9abb553ca40d7171744a8735  -
```

Another interesting thing is that Etherscan seems to be running the same version than the one distributed by Homebrew. It may actually be solcjs. I got into [this other bug](https://github.com/ethereum/solc-js/issues/460) when trying to reproduce it with solcjs.

Any idea what's going on? Are the solc-bin.ethereum.org binaries somehow corrupted? This is really important to us, as it can lead to people deploying contracts that aren't verifiable on Etherscan.

cameel:
~PR branch based on #77. Don't merge until that PR is merged!~ It's on `gh-pages` now.
Fixes https://github.com/ethereum/solidity/issues/10183.

This PR contains macOS binaries in versions between 0.3.6 and 0.6.0 rebuilt with the change from #75. This fixes the discrepancy in the generated bytecode.

~Old binaries are preserved in `macosx-amd64/retired-rebuilds-with-version-discrepancy/`. I'm not sure if it's worth keeping them. They do let people verify stuff they have already built using them but hopefully not many people were affected by that - macOS binaries were added only very recently.~


cameel:
**EDIT 2021-01-24: This comment is now out of date. See updated details further down.**

Current status of the bytecode comparison (from #77, which is still WIP): https://github.com/cameel/solc-bin/pull/1/checks?check_run_id=1594500955

Here's the relevant part of the output: [bytecode-comparison-2020-12-22.txt](https://github.com/ethereum/solc-bin/files/5861343/bytecode-comparison-2020-12-22.txt)

There are still 13 versions with differences (and 0.4.11 which was skipped in this run). From what I've seen so far, it's likely that `prepare_report.py` just needs more workarounds for the older compilers but we'll see.

I have already detected one binary with bytecode differences, though it's a Linux one: `solc-linux-amd64-v0.4.10+commit.f0d539ae` actually identifies itself as `solc-linux-amd64-v0.4.10+commit.f0d539ae`.

cameel:
I'm done verifying the binaries. Unfortunately there are a few versions with problems that prevent the PR check from #77 from passing. This is, however, all caused by factors other than the macOS rebuild - bytecode differences from binaries on other platforms and macOS differences that can't really be fixed. See below for details.

### Report differences
- Emscripten binaries for version 0.3.6 were not built from the exact commit which is tagged as `v0.3.6` so they have a different version string and produce different bytecode than the macOS binary. See https://github.com/ethereum/solidity/issues/10846.
- In 0.4.7 and 0.4.8 the bytecode is different on every platform and there's not much that can be done about it. See https://github.com/ethereum/solidity/issues/10841.
- On 0.4.10 and 0.4.26 it's the Linux binaries that fail the comparison. See https://github.com/ethereum/solidity/issues/10839 and https://github.com/ethereum/solidity/issues/10840.
- 0.4.11 seemingly fails comparison on all platforms but that's only because the results get diffed with the wasm report, which is the broken one in this case.
    - The problem is that there are some LLL test cases that crash the emscripten versions of the compiler when passed in as Solidity code. Example of a test case that causes a crash: `(returnlll (return (exp (- 0 2) 3)))`. This does not happen with native binaries.
    - I have added a try/catch block in `prepare_report.js` to handle the crash but unfortunately the compiler prints an exit code to stdout before it crashes. This throws off the comparison. If you download the reports and manually remove all lines that consist only of the number `5` or `26`, the content becomes the same on all platforms. There are no actual differences in the bytecode.
        - I could work around it by rewriting `prepare_report.js` to write to file (rather than to stdout) but I'm not sure it's worth the extra effort just for this single problematic case.

In all other cases the reports are identical on all platforms (except Windows which I could not check). Here's the most relevant part of the output of the [last run of the `compare` job](https://github.com/ethereum/solc-bin/pull/76/checks?check_run_id=1761261024) if you want to see the details: [bytecode-comparison-2021-01-25.txt](https://github.com/ethereum/solc-bin/files/5866087/bytecode-comparison-2021-01-25.txt).

To make sure the binaries I did not replace are fine too, I have created a PR in my fork that touched all the binaries to force a [CI run covering all of them](https://github.com/cameel/solc-bin/pull/2/checks?check_run_id=1761370014). Here are the results: [bytecode-comparison-all-2021-01-25.txt](https://github.com/ethereum/solc-bin/files/5866088/bytecode-comparison-all-2021-01-25.txt). There are no extra differences.

cameel:
Also, I decided not to leave the buggy builds in a subdirectory. They should not be used in normal circumstances and they're still available in repo history in case anyone needs them for verification anyway. I have created a tag to make finding them easier: [`macosx-binaries-reporting-wrong-versions`](https://github.com/ethereum/solc-bin/releases/tag/macosx-binaries-reporting-wrong-versions).

cameel:
We have decided to close https://github.com/ethereum/solidity/issues/10841 and leave the 0.4.7 and 0.4.8 binaries in the repo.

In that case should I hard-code the PR check to skip these versions? This is an option but it might also be misleading (by not being apparent that they were skipped) and we're not likely to modify these binaries in the future anyway.

cameel:
Rebased on `gh-pages` with conflicts resolved.
Let's make sure it passes the checks though.

axic:
Did we end up releasing the blog post about these?

PR #80, opened 2021-01-28
Replace linux 0.4.26 build with the binary from 0.4.26 docker image

cameel:
Fixes https://github.com/ethereum/solidity/issues/10840.
~Depends on #76.~
Depends on #91.

I'm not sure if we really want to solve https://github.com/ethereum/solidity/issues/10840 this way but it's certainly one way to do this.
The binary from the [0.4.26 docker image](https://hub.docker.com/layers/ethereum/solc/0.4.26/images/sha256-2e9558c26d9b8659e52fea33a6f744190c5ad9ac8a70e628097ab2fbdc30800e) produces different bytecode on the problematic test case and the asm version looks like the one produced by binaries for other platforms. I used it to replace the one currently in this repo. I'm creating this PR as a draft to see if it passes the comparison on all test cases.

If we want this then we should consider also replacing the binary on the github release page (it's identical to the one in solc-bin).

PR #81, opened 2021-02-09
add sha256 hash to the lists

### LINKED ISSUE #10379: Add SHA-256 hashes to file lists in solc-bin
Proposed by @axic in https://github.com/ethereum/solidity/pull/10352#discussion_r527268367.

> Hmm, didn't we also had a sh256 hash next to this? That is way more useful for anyone downloading a static compiler from here and trying to verify it on CI (as that works with shasum -a 256 <file>).

After doing this the example in solc-bin docs should be updated to include such a hash too.

ssi91:
Fixes https://github.com/ethereum/solidity/issues/10379

Relevant chances in the docs can be found there https://github.com/ethereum/solidity/pull/10932

### DIFF
--- a/update
+++ b/update
@@ -134,11 +134,12 @@ async function makeEntry (dir, parsedFileName, oldList) {
     }
   }
 
-  if (!build.keccak256 || !build.urls || build.urls.length !== 2) {
+  if (!build.sha256 || !build.keccak256 || !build.urls || build.urls.length !== 2) {
     const readFile = util.promisify(fs.readFile)
     const fileContent = await readFile(absolutePath)
     build.keccak256 = '0x' + ethUtil.keccak(fileContent).toString('hex')
     console.log("Computing hashes of '" + pathRelativeToRoot + "'")
+    build.sha256 = '0x' + ethUtil.sha256(fileContent).toString('hex')
     build.urls = [
       'bzzr://' + swarmhash(fileContent).toString('hex'),
       'dweb:/ipfs/' + await ipfsHash(fileContent)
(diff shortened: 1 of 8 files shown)


update
```diff
@@ -139,6 +139,7 @@ async function makeEntry (dir, parsedFileName, oldList) {
     const fileContent = await readFile(absolutePath)
     build.keccak256 = '0x' + ethUtil.keccak(fileContent).toString('hex')
     console.log("Computing hashes of '" + pathRelativeToRoot + "'")
+    build.sha256 = '0x' + ethUtil.sha256(fileContent).toString('hex')
```
chriseth, line 142: Can you also modify line 137 accordingly, please?

ssi91: @chriseth done. Mark as resolved if it's well now :)

cameel (dismissed):
@chriseth Looks like the lint/update PR check did not run on this PR. And it's not just that it's not shown in the PR - I can't see it on CircleCI either: https://app.circleci.com/pipelines/github/ethereum/solc-bin. I wonder why. It does run on my PRs (see for example #80). Maybe it's something about permissions?

The PR itself seems fine but I'm marking it as "changes requested" to prevent merging for now - I think that PR is very important in this case.

cameel (approved):

cameel:
Looks like the check did not run because this is a PR from an external repo. @axic has made the check mandatory now so at least we'll be able to see that it's missing but for security reasons we don't want to allow external PRs to run it.

@ssi91 I have resubmitted your PR as #82 and I'm closing this one.

PR #82, opened 2021-02-15
add sha256 hash to the lists

### LINKED ISSUE #10379: Add SHA-256 hashes to file lists in solc-bin

cameel:
Fixes ethereum/solidity#10379.

I'm just re-submitting #81 by @ssi91 as an internal PR so that it can pass all the PR checks.

### DIFF
--- a/bin/list.json
+++ b/bin/list.json
@@ -6,6 +6,7 @@
       "build": "commit.6ff4cd6",
       "longVersion": "0.1.1+commit.6ff4cd6",
       "keccak256": "0xd8b8c64f4e9de41e6604e6ac30274eff5b80f831f8534f0ad85ec0aff466bb25",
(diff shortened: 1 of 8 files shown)


cameel:
Good thing that we have that check actually. I rebased it on top of `gh-pages` and it now fails. And for a good reason - I forgot to rerun it to add sha256 checksums for the nightlies that got added in the meantime.

I'm re-running the update now.

chriseth (approved):

PR #85, opened 2021-03-12
README.md -> README.html on S3 sync

cameel:
A bare-bones markdown rendering for README.md. Not very pretty but still better than serving users a file for download.

### DIFF
--- a/.github/workflows/s3-mirror.yml
+++ b/.github/workflows/s3-mirror.yml
@@ -36,6 +36,11 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Render README.md to HTML
+        run: |
+          pip install markdown
+          echo "<html><body>$(python -m markdown README.md)</body></html>" > README.html
+
       - name: Sync the S3 bucket
         run: |
           ./sync-s3.sh "$S3_BUCKET"


ekpyron (approved):

.github/workflows/s3-mirror.yml
```diff
@@ -36,6 +36,11 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Render README.md to HTML
+        run: |
+          pip install markdown
+          echo "<html><body>$(python -m markdown README.md)</body></html>" > README.html
```
axic, line 42: Hmm, why doesn't that markdown package do the "html/body" wrapping, seems weird?

cameel: Well, its primary use case is being a library not a command-line tool, so I guess you would normally just embed it in a page you already have. The library does have some command-line options but I could not see anything to add the document wrapper around it. I tried `mistune` too but it produced very similar output, without the wrapper (at least by default).

I didn't want to dig too deep into that because I expect that we'll just replace it completely when we add the system that automatically lists available binaries. But if you think it's worthwhile I could try some other libs. For example [markdown-html-action](https://github.com/ZacJW/markdown-html-action) does not add these on its own (it only adds `doctype`) so apparently the library it uses does it automatically.

cameel: Well, its primary use case is being a library not a command-line tool, so I guess you would normally just embed it in a page you already have. The library does have some command-line options but I could not see anything to add the document wrapper around it. I tried `mistune` too but it produced very similar output, without the wrapper (at least by default).

I didn't want to dig too deep into that because I expect that we'll just replace it completely when we add the system that automatically lists available binaries. But if you think it's worthwhile I could try some other libs. For example [markdown-html-action](https://github.com/ZacJW/markdown-html-action) does not add these on its own (it only adds `doctype`) so apparently the library it uses does it automatically.

axic: The documentation (https://python-markdown.github.io/cli/) just mentions `echo "Some **Markdown** text." | python -m markdown > output.html`, without the need for extra html/body wrapping.

cameel: Yeah. Would be nice if they mentioned it explicitly but anyway I found https://github.com/Python-Markdown/markdown/issues/492 which says that it's by design and that a CLI option won't be added because it's out of scope. Unfortunately it's our responsibility to provide a template/wrapper when using it.

axic: Thanks! They could have just documented this on their usage guide, sigh.

PR #86, opened 2021-03-22
Free up disk space in Github Actions

cameel:
Currently all of the Github Actions in this repo that check out a working copy run out of disk space and fail.

Apparently the disk in the Github Actions is big enough (~84 GB) but only ~16 GB is free. The rest is taken up by software. Looks like a lot of it can be recovered though. The two biggest space hogs are .NET and Android SDK. They take ~40 GB total. Unfortunately they have lots of files and subdirectories (.NET has 40k subdirs) and removing them all takes over 10 min.

So far I have determined that removing the following subdirectories frees up ~16 GB and takes only 10-15 s:
```bash
rm -rf /usr/share/swift/                        # 1.3 Gb in 80 subdirs
rm -rf /usr/local/lib/android/sdk/build-tools   # 2.1 Gb in 450 subdirs
rm -rf /usr/share/dotnet/shared                 # 5.3 Gb in 350 subdirs
rm -rf /usr/local/lib/android/sdk/ndk           # 7.6 Gb in 1500 subdirs
```
We could cut the time in half by dropping `ndk` but I think that 15 s is quick enough.

This seems to be necessary only on `ubuntu-latest`. On `macosx-10.15` there's a lot of free space already:
```
 Filesystem      Size   Used  Avail Capacity iused      ifree %iused  Mounted on
/dev/disk1s5   380Gi   11Gi   71Gi    13%  488309 3982052091    0%   /
devfs          184Ki  184Ki    0Bi   100%     636          0  100%   /dev
/dev/disk1s1   380Gi  298Gi   71Gi    81% 7827519 3974712881    0%   /System/Volumes/Data
/dev/disk1s4   380Gi  1.0Mi   71Gi     1%       1 3982540399    0%   /private/var/vm
map auto_home    0Bi    0Bi    0Bi   100%       0          0  100%   /System/Volumes/Data/home
```

### DIFF
--- a/.github/workflows/nightly-emscripten.yml
+++ b/.github/workflows/nightly-emscripten.yml
@@ -20,6 +20,14 @@ jobs:
       nightly-already-exists: ${{ env.NIGHTLY_ALREADY_EXISTS }}
 
     steps:
+      - name: Remove unused software to free up space
+        run: |
+          # These are quick to delete and large enough to make a difference
+          rm -rf /usr/share/swift/                         # 1.3 GB in 80 subdirs
+          rm -rf /usr/local/lib/android/sdk/build-tools/   # 2.1 GB in 450 subdirs
+          rm -rf /usr/share/dotnet/shared/                 # 5.3 GB in 350 subdirs
+          rm -rf /usr/local/lib/android/sdk/ndk/           # 7.6 GB in 1500 subdirs
+
       - uses: actions/checkout@v2
         with:
           repository: 'ethereum/solidity'
@@ -111,6 +119,14 @@ jobs:
 
     if: "needs.build-emscripten-nightly.outputs.nightly-already-exists == 'false'"
     steps:
+      - name: Remove unused software to free up space
+        run: |
+          # These are quick to delete and large enough to make a difference
+          rm -rf /usr/share/swift/                         # 1.3 GB in 80 subdirs
+          rm -rf /usr/local/lib/android/sdk/build-tools/   # 2.1 GB in 450 subdirs
+          rm -rf /usr/share/dotnet/shared/                 # 5.3 GB in 350 subdirs
+          rm -rf /usr/local/lib/android/sdk/ndk/           # 7.6 GB in 1500 subdirs
+
       - uses: actions/checkout@v2
         with:
           ref: ${{ env.TARGET_BRANCH }}
--- a/.github/workflows/t-bytecode-compare.yml
+++ b/.github/workflows/t-bytecode-compare.yml
@@ -27,6 +27,16 @@ jobs:
       PLATFORM: ${{ matrix.platform }}
 
     steps:
+      - name: Remove unused software to free up space
+        # macOS runners apparently have > 70 GB free so it's only needed on Ubuntu
+        if: "env.PLATFORM != 'macosx-amd64'"
+        run: |
+          # These are quick to delete and large enough to make a difference
+          rm -rf /usr/share/swift/                         # 1.3 GB in 80 subdirs
+          rm -rf /usr/local/lib/android/sdk/build-tools/   # 2.1 GB in 450 subdirs
+          rm -rf /usr/share/dotnet/shared/                 # 5.3 GB in 350 subdirs
+          rm -rf /usr/local/lib/android/sdk/ndk/           # 7.6 GB in 1500 subdirs
+
       - uses: actions/setup-python@v2
         with:
           # Use the latest minor release of Python 3. prepare_report.py requires Python >= 3.7
--- a/.github/workflows/random-macosx-build.yml
+++ b/.github/workflows/random-macosx-build.yml
@@ -358,6 +358,14 @@ jobs:
 
     if: "needs.select-solc-version.outputs.solidity-version"
     steps:
+      - name: Remove unused software to free up space
+        run: |
+          # These are quick to delete and large enough to make a difference
+          rm -rf /usr/share/swift/                         # 1.3 GB in 80 subdirs
+          rm -rf /usr/local/lib/android/sdk/build-tools/   # 2.1 GB in 450 subdirs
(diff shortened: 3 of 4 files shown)


.github/workflows/nightly-emscripten.yml
```diff
@@ -20,6 +20,14 @@ jobs:
       nightly-already-exists: ${{ env.NIGHTLY_ALREADY_EXISTS }}
 
     steps:
+      - name: Remove unused software to free up space
+        run: |
+          # These are quick to delete and large enough to make a difference
+          rm -rf /usr/share/swift/                        # 1.3 Gb in 80 subdirs
+          rm -rf /usr/local/lib/android/sdk/build-tools   # 2.1 Gb in 450 subdirs
+          rm -rf /usr/share/dotnet/shared                 # 5.3 Gb in 350 subdirs
+          rm -rf /usr/local/lib/android/sdk/ndk           # 7.6 Gb in 1500 subdirs
```
cameel: I'd prefer to put it in a reusable script instead of repeating it in all actions but unfortunately then I could only access it after I already have a checkout of the repository.

I could retrieve it with `curl` via github API but I'm not sure if complicating it that way is worth it.

cameel:
Another way to reduce disk usage that I've seen others do is to disable swap and remove `/swapfile`.
Also, there are apparently some docker images/containers on the local filesystem that could be removed.

.github/workflows/t-bytecode-compare.yml
```diff
@@ -27,6 +27,16 @@ jobs:
       PLATFORM: ${{ matrix.platform }}
 
     steps:
+      - name: Remove unused software to free up space
+        # macOS runners apparently have > 70 GB free so it's only needed on Ubuntu
+        if: "env.PLATFORM != 'macosx-amd64'"
```
hrkrshnn, line 32: Why do you have this extra check?

hrkrshnn (approved):
LGTM

.github/workflows/t-bytecode-compare.yml
hrkrshnn: Ah, I'm just blind.

hrkrshnn: But it doesn't hurt removing it, does it?

cameel: Yeah. Technically, I could leave it on because it does not fail. But it also does not really remove anything (the files are probably not there, though `/usr/` itself exists and has some stuff in it) so I think it's better to make that clear.

.github/workflows/nightly-emscripten.yml
christianparpart: @cameel you can put that into an action and just make use of it here, that gives you less copy'n'pasting and accidentally forgetting something and a central way to improve the script without touching it in every used repository.

cameel:
In case we want to recover more space later, here's where the most of the big stuff in the `ubuntu-latest` image is:

```
50G /usr/
    16G local/
        1.1G graalvm/
        12G lib/android/
            12G sdk/
                7.6G ndk/
                2.1G build-tools/
    5.4G lib/
    28G  share/
        25G dotnet/
            5.3G shared/
                3.3G Microsoft.NETCore.App/
                1.7G Microsoft.AspNetCore.App/
            19G sdk/
                1.3G NuGetFallbackFolder/
        1.3G swift/
8.0G /opt/
    1.9G ghc/
    4.4G hostedtoolcache/
        1.1G go/
        1.2G CodeQL/
        1.3G Python/
2.2G /var/
    2.1G lib/
4.1G /mnt/
2.4G /snap/

TOTAL 68G
```

And here are some select directories whose removal I have timed:

| Directory                                           | Removal time | Size   | Space recovered | Subdirectories | Errors                                           |
|-----------------------------------------------------|--------------|--------|-----------------|----------------|--------------------------------------------------|
| `/usr/local/graalvm/`                               | 10 s         | 1.1 GB | -2 MB           | 1.6k           | ~5.9k x `Permission Denied`                      |
| `/usr/local/lib/android/`                           | 5 min        | 12 GB  | 11.83 GB        | 18k            | `/usr/local/lib/android/sdk: Permission denied`  |
| `/usr/local/lib/android/sdk/ndk`                    | 21 s         | 7.6 GB | 7.87 GB         | 1.5k           |                                                  |
| `/usr/local/lib/android/sdk/build-tools`            | 3 s          | 2.1 GB | 2.21 GB         | 450            |                                                  |
| `/usr/share/dotnet/`                                | 7 min        | 25 GB  | 25.25 GB        | 41k            |                                                  |
| `/usr/share/dotnet/shared/Microsoft.NETCore.App`    | 1s           | 3.3 GB | 3.36 GB         | 47             |                                                  |
| `/usr/share/dotnet/shared`                          | 7s           | 5.3 GB | 5.56 GB         | 350            |                                                  |
| `/usr/share/dotnet/sdk/NuGetFallbackFolder`         | 2 min        | 1.3 GB | 1.28 GB         | 9.5k           |                                                  |
| `/usr/share/swift/`                                 | 1 s          | 1.3 GB | 1.32 GB         | 78             |                                                  |

The time includes only the removal step, not the whole job that ran it. There's a lot of variability though. For example `/usr/local/lib/android/sdk/ndk` took 22 s on its own but it usually takes about 15 s in my other runs that remove it together with the three other directories.

.github/workflows/nightly-emscripten.yml
cameel: You mean create a reusable action? Not a bad idea. I'll try that.

cameel: i just tried that and unfortunately it won't work without putting the action in a separate repo. The problem is that actions from the same repo do not get checked out automatically and to check it out you need to check out the whole repo which is too big in our case.

chriseth (approved):

PR #87, opened 2021-03-23
Binaries for 0.8.3.

chriseth:


bin/list.json
```diff
@@ -6,215 +6,215 @@
       "build": "commit.6ff4cd6",
       "longVersion": "0.1.1+commit.6ff4cd6",
       "keccak256": "0xd8b8c64f4e9de41e6604e6ac30274eff5b80f831f8534f0ad85ec0aff466bb25",
-      "sha256": "0xfc54135edfe07d835216f87fa6c3b9049e1415c2825027b926daad018e09269f",
       "urls": [
         "bzzr://8f3c028825a1b72645f46920b67dca9432a87fc37a8940a2b2ce1dd6ddc2e29b",
         "dweb:/ipfs/QmPPGxsMtQSEUt9hn9VAtrtBTa1u9S5KF1myw78epNNFkx"
-      ]
+      ],
+      "sha256": "0xfc54135edfe07d835216f87fa6c3b9049e1415c2825027b926daad018e09269f"
```
cameel: Why isn't it sorting the keys alphabetically?

chriseth: I don't know. It also always re-computes the hashes for me. Maybe an older npm version?

chriseth: Acutally it never has, has it? `keccak256` should be before `longVersion`.

chriseth: let me look into at leat getting a minimal diff.

PR #88, opened 2021-03-23
Fix taking sha256 from previous list.

chriseth:


update
```diff
@@ -129,6 +129,7 @@ async function makeEntry (dir, parsedFileName, oldList) {
         throw Error("Found multiple list.json entries for binary '" + pathRelativeToRoot + "'")
       } else if (entries.length === 1) {
         build.keccak256 = entries[0].keccak256
+        build.sha256 = entries[0].sha256
```
chriseth, line 132: It is actually not sorted alphabetically, but sorted by order of assignment...

cameel: ok. well, as long as it's deterministic, it's fine I guess.

cameel (approved):

PR #92, opened 2021-05-10
Replace the bad Windows binary for 0.8.2 with one from github release page

### LINKED ISSUE #11358: Windows build of 0.8.2 seems to lack 0.8.2 features
I reported this in the chat but I want to leave it here because there may be something wrong in the build pipeline or release process that should probably be looked into.

The Windows build for 0.8.2 doesn't seem to support `@custom` NatSpec tags. Later builds (0.8.3, 0.8.4) do not have this problem.

Reproduction: https://github.com/frangio/solidity-0.8.2-windows

cameel:
Depends on #91.
Fixes  https://github.com/ethereum/solidity/issues/11358.

I have also compared all binaries between 0.8.0 and 0.8.4 with the ones on the github release page and this is the only one that differed.

### DIFF
--- a/windows-amd64/list.json
+++ b/windows-amd64/list.json
@@ -773,11 +773,11 @@
       "version": "0.8.2",
       "build": "commit.661d1103",
       "longVersion": "0.8.2+commit.661d1103",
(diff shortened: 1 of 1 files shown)


axic:
Did we end up releasing the blog post about these?

PR #97, opened 2021-09-06
[s3-mirror] Invalidate cached lists and some symlinks after sync

### LINKED ISSUE #11572: Too aggressive caching for `soljson-latest.js` in solc-bin
This came up in https://github.com/juanfranblanco/vscode-solidity/issues/256#issuecomment-866259220.

Some people are getting a stale version of `soljson-latest.js` from https://binaries.soliditylang.org. It's 0.8.5 rather than 0.8.6.

The file is served from an S3 bucket with a Cloudfront CDN in front of it (the config is described in https://github.com/ethereum/solc-bin/pull/47#issuecomment-666447083). Binaries are cached because they are supposed to never change but file lists (i.e. any file matching `*/.list.*`) are exempt from caching.

We should probably exempt `soljson-latest.js` and other files that may change content as well. The problem is that these files do get a lot of traffic. For example `bin/soljson-latest.js` got over 200k requests in the last month (98.93% out of which came from cache;  only 1350 were misses). In total 5 TB of traffic for this file alone, all but 30 GB coming from cache. This is the top downloaded file (though a few others are close). So before doing that we need to check how much it will affect the cost. I think the impact might not be that big, since serving from an S3 bucket should be relatively cheap anyway but it's better to check. If the impact turns out to be too big, we might consider invalidating cache for specific files after every update, in the Github Action responsible for mirroring solc-bin instead of disabling cache for all accesses to the file.

cameel:
Fixes https://github.com/ethereum/solidity/issues/11572
This is also the second part of the fix for https://github.com/ethereum/devops/issues/789

The first part was to re-enable caching for list files. This should improve transfer speeds because people will be getting them from CloudFront cache rather than from our regional S3 bucket.

The downside is that changes to them might not be visible for up to 24h after we sync. The second part of the fix helps with that - we issue a cache invalidation command immediately after the sync.

### DIFF
--- a/sync-s3.sh
+++ b/sync-s3.sh
@@ -27,7 +27,8 @@ set -eo pipefail
 die() { >&2 echo "ERROR: $@" && false; }
 
 s3_bucket="$1"
-(( $# == 1 )) || die "Expected exactly 1 parameter."
+cloudfront_distribution_id="$2"
+(( $# == 2 )) || die "Expected exactly 2 parameters."
 
 [[ $(git rev-parse --is-shallow-repository) == false ]] || die "This script requires access to full git history to be able to set file timestamps correctly."
 
@@ -59,3 +60,17 @@ aws s3 sync . "s3://${s3_bucket}" --delete --follow-symlinks --no-progress --exc
 
 echo "===> Syncing file lists with the S3 bucket"
 aws s3 sync . "s3://${s3_bucket}" --delete --follow-symlinks --no-progress --exclude "*" --include "*/list.*"
+
+echo "===> Invalidating CloudFront cache"
+# Invalidate only the files that might change in-place when new binaries are added.
+# NOTE: Invalidation paths allow wildcards only as the last character. When used at
+# any other position, AWS will not report an error but will also not invalidate it.
+# NOTE: The code below assumes that paths do not contain whitespace.
+aws cloudfront create-invalidation \
+    --distribution-id "$cloudfront_distribution_id" \
+    --paths \
+        /bin/soljson-nightly.js \
+        /soljson.js \
+        $(find . -wholename '*/list.*' | cut --characters 2-) \
+        $(find . -wholename '*/*-latest' | cut --characters 2-) \
+        $(find . -wholename '*/*-latest.*' | cut --characters 2-)
--- a/sync-s3.sh
+++ b/sync-s3.sh
@@ -44,8 +44,8 @@ done
 echo "===> Removing files that should not be uploaded to S3"
 # NOTE: This ensures that they will be deleted from the bucket if they're already there.
 # If we used `aws s3 sync --delete --exclude` instead, they would not get deleted.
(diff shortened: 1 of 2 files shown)


cameel:
CI is failing here because of the recent changes in solc-js. Fix: https://github.com/ethereum/solidity/pull/11904.

cameel:
Looks like for some reason these wildcards do not invalidate `bin/list.json` (and maybe other files but for this one I have tested it). Only direct invalidation of `/bin/list.json` does it. I need to investigate so switching to draft.

cameel:
ok, here's my answer:

[Invalidation paths](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Invalidation.html#invalidation-specifying-objects-paths)
> You can also invalidate multiple files simultaneously by using the `*` wildcard. The `*`, which replaces 0 or more characters, must be the last character in the invalidation path.

So my patterns are wrong after all. I thought they were valid because AWS does not report an error here while it does when I, for example, use wildcards wrong in CORS config in S3 or if I don't use a leading `/` in an invalidation path.

sync-s3.sh
```diff
@@ -59,3 +59,17 @@ aws s3 sync . "s3://${s3_bucket}" --delete --follow-symlinks --no-progress --exc
...
 echo "===> Syncing file lists with the S3 bucket"
 aws s3 sync . "s3://${s3_bucket}" --delete --follow-symlinks --no-progress --exclude "*" --include "*/list.*"
+
+echo "===> Invalidating CloudFront cache"
+# Invalidate only the files that might change in-place when new binaries are added.
+# NOTE: Invalidation paths allow wildcards only as the last character. When used at
+# any other position, AWS will not report an error but will also not invalidate it.
+# NOTE: The code below assumes that paths do not contain whitespace.
+aws cloudfront create-invalidation \
+    --distribution-id E1O6GT57WUFUHD \
```
chriseth: Where does the ID come from? Is it used multiple times (in other files)?

chriseth (approved):

sync-s3.sh
cameel: It's the ID CloudFront uses to identify the "distribution", i.e. the project you create there to cache and serve stuff from the S3 bucket. We have only one distribution and it's permanent so this should not change unless we scrap it all and configure CloudFront from scratch.

It's not a secret key of any kind if that's what you're worried about. Secrets are in https://github.com/ethereum/solc-bin/blob/6b4ae0223301ecba154b963e0c318308f52a9cee/.github/workflows/s3-mirror.yml#L37-L41 and they come from repo config.

This script is the only place we're using it currently. But you have a point, I see that I did not even hard-code the bucket IDs so having this ID here is a bit out of place. I'll move it to `s3-mirror.yml`.

cameel: Done.

chriseth (approved):

PR #99, opened 2021-09-29
Solidity 0.8.9

hrkrshnn:


chriseth (approved):

chriseth (approved):

hrkrshnn:
Seems that the hashes aren't correct. Weird https://app.circleci.com/pipelines/github/ethereum/solc-bin/844/workflows/f39d6613-26cd-4bc5-b862-ae47596f5be1/jobs/857

hrkrshnn:
I'll try to run `./update` and see if the hashes change. I think the issue was that the second `./update --reuse-hashes` after copying the right `soljson` is wrong, because it shouldn't reuse hashes :(

Edit: done.

PR #100, opened 2021-11-05
Update dependencies to the latest versions

cameel:
Depends on #101 (technically it's independent but tests won't pass without it). Draft until that PR is merged.

While trying to solve the problem from #101 I tried updating dependencies. This did not solve it but we should update them anyway.

One of the dependencies ([js-ipld](https://github.com/ipld/js-ipld)) seems to have been deprecated and replaced by [js-multiformats](https://github.com/multiformats/js-multiformats). We should switch eventually but this PR only updates it to the latest version of the original package for now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,8 @@ version: 2.1
 jobs:
   chk_npm_lint_and_test:
     docker:
-      - image: circleci/node
+      # FIXME: sha3 package fails to build on latest node.js (version 17)
+      - image: circleci/node:16
     resource_class: xlarge
     steps:
       - checkout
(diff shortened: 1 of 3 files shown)


cameel:
Well, looks like this does not solve the problem. Let's see if downgrading node.js does the trick.

cameel:
Looks like node 16 helps. It's still failing only due to new linter checks introduced by `started`. I'm going to create separate PR just with the node.js downgrade.

cameel:
Linter problems resolved and #101 is now merged so I'm making this reviewable.

chriseth (approved):

PR #101, opened 2021-11-05
Downgrade node.js in npm&lint PR check to work around sha3 build failure

cameel:
The [`chk_npm_lint_and_test`](https://app.circleci.com/pipelines/github/ethereum/solc-bin/872/workflows/457e6725-fdba-4168-83b6-f0751c5df92a/jobs/885) job on CircleCI has been failing for a few days now. There seems to be some problem building the `sha3` module:

```
npm ERR! ../../nan/nan_typedarray_contents.h:34:43: error: ‘class v8::ArrayBuffer’ has no member named ‘GetContents’
npm ERR!    34 |       data   = static_cast<char*>(buffer->GetContents().Data()) + byte_offset;
npm ERR!       | 
```

It does not fail for me locally on node.js 14 or 16 but looks like we're running node.js 17 in CI. Downgrading resolves the problem.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,8 @@ version: 2.1
 jobs:
   chk_npm_lint_and_test:
     docker:
-      - image: circleci/node
+      # FIXME: sha3 package fails to build on latest node.js (version 17)
+      - image: circleci/node:16
     resource_class: xlarge
     steps:
       - checkout


ekpyron:
Should probably be reported to https://github.com/phusion/node-sha3/issues, shouldn't it?

ekpyron:
> warning: cast between incompatible function types from ‘void (*)(Nan::ADDON_REGISTER_FUNCTION_ARGS_TYPE)’ {aka ‘void (*)(v8::Local<v8::Object>)’} to ‘node::addon_register_func’ {aka ‘void (*)(v8::Local<v8::Object>, v8::Local<v8::Value>, void*)’} [-Wcast-function-type]

One really shouldn't look too closely at these things, if one wants to stay sane...

ekpyron (approved):
Well, I guess downgrading is an ok option for now.

cameel:
Probably. I wasn't sure if we're actually using latest sha3 here and locally I don't have node.js 17 to test it with latest deps. #100 still failed but it's only a transitive dependency so it still could be something fixed upstream.

Oh, well, I'll report it anyway and see what they say.

ekpyron:
This looks like it's some hidden dependency of this sha-3 thing on "nan"? Which actually had this fixed one-and-a-half years  ago in https://github.com/nodejs/nan/commit/2c023bd447661a61071da318b0ff4003c3858d39 ?
Stuff like this is what you get if people use package managers everywhere...

ekpyron:
> Probably. I wasn't sure if we're actually using latest sha3 here and locally I don't have node.js 17 to test it with latest deps. #100 still failed but it's only a transitive dependency so it still could be something fixed upstream.
> 
> Oh, well, I'll report it anyway and see what they say.

Well, if we can't say where a potentially outdated dependency is pulled in, it's of course hard to reasonably report anything. It looks to me like this is *actually* fixed at the source in the `nan` thing, but who knows what weird dependency chains lead there...

cameel:
Looks like the dependency on sha3 is indeed outdated and it gets pulled in by keccakjs. [It's `^1.2.2` there](https://github.com/axic/keccakjs/blob/d3bc1af663aa3525efd05b152765f2c8950494b2/package.json#L23) while the latest is 2.1.4. Looks like it's @axic's package though so he could just update his deps.

cameel:
I reported it in https://github.com/axic/keccakjs/issues/13 to give @axic a reminder to do it eventually. Looks like https://github.com/axic/keccakjs/issues/10 would be an alternative solution because it would let us keep the requirement on our side.

PR #105, opened 2021-12-15
Reduce concurrency in the `update` script to lower memory usage

### LINKED ISSUE #9956: Nightly builds run out of memory
I have just noticed that the nightly build action in [solc-bin](https://github.com/ethereum/solc-bin) has been failing for three days in a row now: [1](https://github.com/ethereum/solc-bin/runs/1201552073?check_suite_focus=true), [2](https://github.com/ethereum/solc-bin/runs/1204181256?check_suite_focus=true), [3](https://github.com/ethereum/solc-bin/runs/1206816217?check_suite_focus=true).

The output suggests that it's being killed by github.

I suspect it's either running out of memory or out of time and the former seems more likely. I just ran `./update` locally and RAM usage peaks at 11.4 GB. It's somewhere around 10 GB right about the point where it fails on github.

### The error
```
Computing hashes of '/bin/soljson-v0.7.3-nightly.2020.9.28+commit.dd5b0a71.js'
Computing hashes of '/bin/soljson-v0.7.3-nightly.2020.9.29+commit.343c13f9.js'
Computing hashes of '/bin/soljson-v0.7.3-nightly.2020.9.30+commit.3af21c92.js'
Killed
npm ERR! code ELIFECYCLE
npm ERR! errno 137
npm ERR! solc-bin@0.0.0 update: `./update`
npm ERR! Exit status 137
npm ERR! 
npm ERR! Failed at the solc-bin@0.0.0 update script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/runner/.npm/_logs/2020-10-05T01_05_31_133Z-debug.log
Error: Process completed with exit code 137.
```

cameel:
Fixes https://github.com/ethereum/solidity/issues/9956.

This turned out to be pretty easy to solve without extra libraries, I just needed to read up a bit on the specifics of JavaScript's `async`/`await` mechanism and promises. A lot of the noise in the PR really comes from small refactors.

This adds an option called `--max-files-per-batch` that lets you specify how many files to process simultaneously. The number is actually the number of async calls per batch. Each call processes a single binary and there are multiple batches being processed simultaneously, but only one per directory. So by setting it to N you'll get D * N simultaneous calls, where D is the number of dirs containing binaries.

After testing with various limits to figure out what to use in CI it looks like we might not be losing anything by setting it simply to 1:
- `--max-files-per-batch 1`: time (user+sys) = 533+13s, memory = ~350 MB
- `--max-files-per-batch 50`: time (user+sys) = 523+10s, memory = ~1.5 GB
- no limit: time (user+sys) = 531+10s, memory = ~11 GB MB

Surprisingly, looks like there's virtually no benefit from async processing. Maybe the time spent being blocked on disk I/O is negligible compared to time spent hashing?

Since we were using an `xlarge` machine only due to memory usage, I'm changing that to `small` now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   },
   "scripts": {
     "lint": "standard update",
-    "update": "./update",
+    "update": "./update --max-files-per-batch 1",
     "test": "git checkout -f && npm run update && git diff --exit-code"
   },
   "repository": {
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ jobs:
     docker:
       # FIXME: sha3 package fails to build on latest node.js (version 17)
       - image: circleci/node:16
-    resource_class: xlarge
+    resource_class: small
     steps:
       - checkout
       - run: npm install
--- a/update
+++ b/update
@@ -155,6 +155,18 @@ async function makeEntry (dir, parsedFileName, oldList) {
   return build
 }
 
+async function batchedAsyncMap (values, batchSize, asyncMapFunction) {
+  if (batchSize === null) {
+    batchSize = values.length
+  }
+
+  let results = []
+  for (let i = 0; i < values.length; i += batchSize) {
+    results = results.concat(await Promise.all(values.slice(i, i + batchSize).map(asyncMapFunction)))
+  }
+  return results
+}
+
 function processDir (dir, options, listCallback) {
   fs.readdir(path.join(__dirname, dir), { withFileTypes: true }, async function (err, files) {
     if (err) {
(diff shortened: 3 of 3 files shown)


cameel:
CI running time seems largely unaffected. I compared with the most recent release PR (#102) and in both cases the update took ~10 min.

package.json
```diff
@@ -16,7 +16,7 @@
   },
   "scripts": {
     "lint": "standard update",
-    "update": "./update",
+    "update": "./update --max-files-per-batch 1",
```
chriseth, line 19: Is 1 really the right number? Shouldn't we balance out IO and hash computations?

chriseth: Ah you wrote that in the issue description. Maybe it's because the files were recently written to disk and are still in the cache or something...

chriseth (approved):

package.json
cameel: I don't think so - at least in CI the running time did not change significantly. I think that reading files from disk just does not take that much time compared to hashing, at least on a fast disk. This is also why it's so much faster without `--update-hashes`.

In any case, by default there's no limit so, if there is any speed decrease, you won't get it when running it locally.

PR #108, opened 2022-01-30
Removes js-ipld dependency and switches with js-multiformats

elijahhampton:
Replaces ipld-block Block class with multiformats/block.
Replaces ipld CID with multiformats/cid


package-lock.json
```diff
@@ -0,0 +1,3997 @@
+{
+  "name": "solc-bin",
+  "version": "0.0.0",
+  "lockfileVersion": 1,
```
cameel: Please do not include the lockfile in the PR. We don't want to lock to specific versions.

update
```diff
@@ -1,32 +1,42 @@
...
-const semver = require('semver')
-const ethUtil = require('ethereumjs-util')
-const ipfsImporter = require('ipfs-unixfs-importer')
-const IPLD = require('ipld')
-const inMemory = require('ipld-in-memory')
-const swarmhash = require('swarmhash')
+"use strict";
+
+const fs = require("fs");
+const util = require("util");
```
cameel: Can you remove any stylistic changes from this file? It's pretty much impossible to review like this.

If you really want to reformat it, you can submit a separate PR but if you do this, please ensure that our `npm run lint` command enforces the same style. Otherwise it won't be maintained.

package.json
```diff
@@ -4,12 +4,17 @@
...
   "dependencies": {
-    "semver": "^7.3.5"
+    "ipld-dag-cbor": "^1.0.1",
+    "ipld-dag-pb": "^0.22.3",
+    "ipld-raw": "^7.0.1",
+    "multicodec": "^3.2.1",
+    "multiformats": "^9.6.2",
+    "semver": "^7.3.5",
+    "typical": "^7.1.1"
   },
```
cameel: Aren't they all just dev dependencies?

package-lock.json
```diff
@@ -0,0 +1,3997 @@
+{
```
cameel: Can't comment on it directly so I'll leave a comment here: the `.DS_store` file is your local metadata and we don't want that in the repository. Please remove it from the PR.

cameel (changes requested):
Thanks for the PR!

It needs some cleanup before I can review it properly. See my comments below.

update
```diff
@@ -416,3 +424,375 @@ DIRS.forEach(function (dir) {
     })
   }
 })
+
+class IPLDMock {
```
cameel: That's a lot of code for something that was a single line before. It nearly doubled the size of the script :) And we'd need to check if its license is compatible with ours (GPLv3).

Honestly, I'd rather stay with the old `ipld` than maintain something like this. There must be a way to do the same thing with the new libs that replaced it.

cameel (changes requested):

PR #109, opened 2022-02-03
Test for the bytecode compare job with my solc-js fix

cameel:
This is just to make sure PR checks still work after [`adjust-solc-js-installation`](https://github.com/ethereum/solidity/tree/adjust-solc-js-installation). **Should not be merged.**

### DIFF
--- a/.github/workflows/t-bytecode-compare.yml
+++ b/.github/workflows/t-bytecode-compare.yml
@@ -46,6 +46,7 @@ jobs:
         with:
           repository: 'ethereum/solidity'
           path: 'solidity/'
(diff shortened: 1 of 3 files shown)


cameel:
The script indeed needed some adjustments (see [the failed `generate (emscripten-asmjs, ubuntu-latest)` run](https://github.com/ethereum/solc-bin/runs/5055301399)). I fixed it and now [tests pass](https://github.com/ethereum/solc-bin/runs/5055519119) (except for `chk_npm_lint_and_test` but that's only because I modified binaries for the test). Closing this.

PR #110, opened 2022-02-03
Make `soljson.js` a symlink again

cameel:
#99 replaced `soljson.js` at the root of the repository (which as a symlink) with an actual binary. This is wrong because this way the file is not getting updated when we publish a release (the `update` script ignores it on purpose assuming it's a symlink).

This PR fixes this by restoring the symlink.


cameel:
We should also consider retrying #33 and #39. Now that we are getting binaries from S3 and not from GH pages, I think that symlinks would work just fine because we convert them into copies on the server.

ekpyron (approved):

PR #120, opened 2022-09-08
Binaries for 0.8.17

r0qs:

PR #121, opened 2022-09-08
Binaries for 0.8.17

r0qs:


cameel:
When making a new PR it's always good to link to the old one. Makes it easier to figure out later what happened.

In this case we're replacing #120 because it was made from a fork and CI checks won't pass there (the bytecode diffing script can't construct a proper commit range).

cameel (approved):

PR #123, opened 2022-09-30
Pass prerelease option to nightly builds

r0qs:
Updates the CI job to use `emscripten_build` script with `prerelease` parameter introduced by https://github.com/ethereum/solidity/pull/13581.

Depends on https://github.com/ethereum/solidity/pull/13581.

### DIFF
--- a/.github/workflows/nightly-emscripten.yml
+++ b/.github/workflows/nightly-emscripten.yml
@@ -50,7 +50,6 @@ jobs:
           last_commit_hash=$(git rev-parse --short=8 HEAD)
           solidity_version=$("scripts/get_version.sh")
           nightly_version="v${solidity_version}-nightly.${last_commit_date}+commit.${last_commit_hash}"
-          echo "nightly.${last_commit_date}" > prerelease.txt
 
           echo "LAST_COMMIT_DATE=${last_commit_date}" >> $GITHUB_ENV
           echo "LAST_COMMIT_HASH=${last_commit_hash}" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


cameel:
@r0qs Do you need permissions access to solc-bin? We have same problem as last time, i.e. the PR checks won't pass because this is an external PR.

I actually haven't considered that this will affect all contributions. In that case we might want to fix that after all. See https://github.com/ethereum/solidity/issues/13614.

For now I'll just force-merge this one (once the dependency in the main repo is merged).

cameel (commented):
Looks good but not approving yet since I have not yet reviewed the dependency in the main repo.

r0qs:
> @r0qs Do you need permissions access to solc-bin? We have same problem as last time, i.e. the PR checks won't pass because this is an external PR.

No, I have permission. Sorry, it was my mistake, I set two git remote trackers in my fork and pushed the branch to the wrong one, haha
I just realized that I did the same for the other PR in the solidity repo. Should I recreate the PR?

> I actually haven't considered that this will affect all contributions. In that case we might want to fix that after all. See [ethereum/solidity#13614](https://github.com/ethereum/solidity/issues/13614).

Ah, indeed.

cameel:
> Should I recreate the PR?

Like I said, I can just force merge it so no need.

cameel:
@r0qs This needs to be updated to match https://github.com/ethereum/solidity/pull/13581 (and probably also rebased).

cameel (approved):

PR #124, opened 2022-11-03
Replace deprecated js-ipld module

r0qs:
Fix https://github.com/ethereum/solidity/issues/12244

This PR adds a workaround to switch from the deprecated `js-ipld` to `js-multiformats`. It dynamically loads the ipfs modules that are now ESM-only. The proper fix requires us to migrate our modules from CommonJS to ESM format.

Most of the ipfs modules became [ESM-only](https://github.com/ipfs/js-ipfs/blob/master/docs/upgrading/v0.62-v0.63.md#esm), including `ipfs-unixfs-importer`, which dropped support to CommonJS in version [10.0.0](https://github.com/ipfs/js-ipfs-unixfs/blob/master/packages/ipfs-unixfs-importer/CHANGELOG.md#-breaking-changes-1).

PR #125, opened 2022-11-04
Remove redundant libraries to compute IPFS hash

### LINKED ISSUE #12244: Switch from js-ipld to js-multiformats in solc-bin
While updating dependencies in https://github.com/ethereum/solc-bin/pull/100#issue-774215950 I noticed that [js-ipld](https://github.com/ipld/js-ipld), which we depend on seems to have been deprecated and replaced by [js-multiformats](https://github.com/multiformats/js-multiformats).

The task is simply to change the dependency and figure out what object to pass to `ipfsImporter()` instead of `IPLD` in [`ipfsHash()`](https://github.com/ethereum/solc-bin/blob/876858b2df0ea6f10b62e06485401db75176867c/update#L18-L22).

0xGeorgii:
Fixes https://github.com/ethereum/solidity/issues/12244.

@cameel I have made a research of how to use `multiformat` and how it can be applicable here. It is possible but requires rewriting quite a bunch of code. However, I decided to juggle a bit the code and founded out that here:
https://github.com/ethereum/solc-bin/blob/33fb994d977bbf271685f65b5137c7f547323baf/update#L19
`inMemory` function only append in memory storage to a newly instantiated `IPLD` object.

As you advised, I removed all the artifacts by `rm */list.json */list.txt */list.js` and ran `./update`. The only method from `IPLD` that was used is `constructor`. It looks like without manipulation with block all other methods provided by `IPLD` do not needed. Replacement with a simple `IPLDProvider` works as the previous version did.

If I made a mistake somewhere please advise.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "private": true,
   "name": "solc-bin",
+  "type": "module",
   "version": "0.0.0",
   "description": "Current and historical (emscripten) binaries for Solidity",
   "dependencies": {
@@ -8,15 +9,14 @@
   },
   "devDependencies": {
     "ethereumjs-util": "^7.1.3",
-    "ipfs-unixfs-importer": "^9.0.6",
-    "ipld": "^0.30.2",
-    "ipld-in-memory": "^8.0.0",
+    "blockstore-core": "^2.0.2",
+    "ipfs-unixfs-importer": "^11.0.1",
     "standard": "^16.0.4",
     "swarmhash": "^0.1.1"
   },
   "scripts": {
-    "lint": "standard update",
-    "update": "./update --max-files-per-batch 1",
+    "lint": "standard update.mjs",
+    "update": "node update.mjs --max-files-per-batch 1",
     "test": "git checkout -f && npm run update && git diff --exit-code"
   },
   "repository": {
--- a/update.mjs
+++ b/update.mjs
@@ -1,27 +1,36 @@
-#!/usr/bin/env node
-
-'use strict'
-
-const fs = require('fs')
-const util = require('util')
-const path = require('path')
-const semver = require('semver')
-const ethUtil = require('ethereumjs-util')
-const ipfsImporter = require('ipfs-unixfs-importer')
-const IPLD = require('ipld')
-const inMemory = require('ipld-in-memory')
-const swarmhash = require('swarmhash')
-
 // This script updates the index files list.js and list.txt in the directories containing binaries,
 // as well as the 'latest' and 'nightly' symlinks/files.
 
+import { fileURLToPath } from 'url'
+import { dirname, join } from 'path'
+
+import {
+  readlinkSync,
+  unlinkSync,
+  symlinkSync,
+  readFileSync,
+  writeFile,
+  readdir,
+  stat,
+  lstat
+} from 'fs'
+
+import semver from 'semver'
+import swarmhash from 'swarmhash'
+import { readFile as readFileAsync } from 'node:fs/promises'
+import { keccak, sha256 } from 'ethereumjs-util'
+import { importer } from 'ipfs-unixfs-importer'
+import { MemoryBlockstore } from 'blockstore-core/memory'
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = dirname(__filename)
+
 const ipfsHash = async (content) => {
-  const iterator = ipfsImporter.importer([{ content }], await inMemory(IPLD), { onlyHash: true })
+  const iterator = importer([{ content }], new MemoryBlockstore(), { onlyHash: true })
   const { value, done } = await iterator.next()
   if (done) {
     throw new Error('Failed to calculate an IPFS hash.')
   }
-
   await iterator.return()
   return value.cid.toString()
 }
@@ -40,14 +49,14 @@ if (typeof(module) !== 'undefined')
 }
 
 function updateSymlinkSync (linkPathRelativeToRoot, targetRelativeToLink) {
-  const absoluteLinkPath = path.join(__dirname, linkPathRelativeToRoot)
+  const absoluteLinkPath = join(__dirname, linkPathRelativeToRoot)
   let linkString
 
   try {
-    linkString = fs.readlinkSync(absoluteLinkPath)
+    linkString = readlinkSync(absoluteLinkPath)
 
     if (targetRelativeToLink !== linkString) {
-      fs.unlinkSync(absoluteLinkPath)
+      unlinkSync(absoluteLinkPath)
       console.log('Removed link ' + linkPathRelativeToRoot + ' -> ' + linkString)
     }
   } catch (err) {
@@ -57,29 +66,29 @@ function updateSymlinkSync (linkPathRelativeToRoot, targetRelativeToLink) {
   }
 
   if (targetRelativeToLink !== linkString) {
-    fs.symlinkSync(targetRelativeToLink, absoluteLinkPath, 'file')
+    symlinkSync(targetRelativeToLink, absoluteLinkPath, 'file')
     console.log('Created link ' + linkPathRelativeToRoot + ' -> ' + targetRelativeToLink)
   }
 }
 
 function updateCopy (srcRelativeToRoot, destRelativeToRoot) {
-  fs.readFile(path.join(__dirname, srcRelativeToRoot), function (err, data) {
+  readFileSync(join(__dirname, srcRelativeToRoot), function (err, data) {
     if (err) {
       throw err
     }
 
-    const absoluteDest = path.join(__dirname, destRelativeToRoot)
-    fs.stat(absoluteDest, function (err, stats) {
+    const absoluteDest = join(__dirname, destRelativeToRoot)
+    stat(absoluteDest, function (err, stats) {
       if (err && err.code !== 'ENOENT') {
         throw err
       }
 
       // If the target is a symlink, we want to replace it with a copy rather than overwrite the file it links to
       if (!err && stats.isSymbolicLink()) {
-        fs.unlinkSync(absoluteDest)
+        unlinkSync(absoluteDest)
       }
 
-      fs.writeFile(absoluteDest, data, function (err) {
+      writeFile(absoluteDest, data, function (err) {
         if (err) {
           throw err
         }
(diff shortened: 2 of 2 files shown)


r0qs (commented):
Hi @GeorgePlotnikov thank you very much for your contribution, but I'm afraid this will not be the best approach for this issue.

Although the [ipld-in-memory](https://github.com/ipld/ipld-in-memory) module is not deprecated, it is not actively maintained. The last commit is from two years ago. Also, it depends on [js-ipfs-block-service](https://www.npmjs.com/package/ipfs-block-service), which has been deprecated.

I would instead try to use the latest version of `ipfs-unixfs-importer`. However, as you can see [here](https://github.com/ethereum/solc-bin/pull/124) they recently dropped support to CommonJS, and we would need to migrate to ESM format. This is a bit more work, but if you want to try, we can help you get it done.

0xGeorgii:
hey, @r0qs thanks a lot for the advice and for the background. Yes, I would like to implement it. Am I right that the best approach would implement hash calculation from the scratch using `multiformats` instead of all these deprecated or outdated libraries?

r0qs:
> Am I right that the best approach would implement hash calculation from the scratch using `multiformats` instead of all these deprecated or outdated libraries?

Yes, this is one way to go, but it would require that we upgrade to ESM modules anyway since [js-multiformat](https://github.com/multiformats/js-multiformats/releases/tag/v10.0.0) also dropped support to CommonJS.

I believe the simpler approach would be using [ipfs-unixfs-importer](https://github.com/ipfs/js-ipfs-unixfs/tree/master/packages/ipfs-unixfs-importer#example) (please take a look at their example in the readme). This would also require migrating to ESM modules. However, it would be a much simpler implementation, i.e., basically, what is already implemented [here](https://github.com/ethereum/solc-bin/pull/124/files#diff-2937013f2181810606b2a799b05bda2849f3e369a20982a4138f0e0a55984ce4R20-R23) but replacing the dynamic imports by the proper static imports in ESM format.

For a reference on how to migrate from CommonJS to ESM, please take a look here: https://pawelgrzybek.com/all-you-need-to-know-to-move-from-commonjs-to-ecmascript-modules-esm-in-node-js/

0xGeorgii:
@r0qs please take a look at the updated solution

ipld-utils.js
```diff
@@ -0,0 +1,16 @@
...
+// as well as the 'latest' and 'nightly' symlinks/files.
+const ipfsHash = async (content) => {
+  const iterator = importer([{ content }], new MemoryBlockstore(), { onlyHash: true })
+  const { value, done } = await iterator.next()
+  if (done) {
+    throw new Error('Failed to calculate an IPFS hash.')
+ }
+ await iterator.return()
+ return value.cid.toString()
+}
```
r0qs: I think it is unnecessary to move this function to another file, we can keep it in the same file as before since it will not be used anywhere else.

package.json
```diff
@@ -1,22 +1,22 @@
...
     "ipfs-unixfs-importer": "^9.0.6",
-    "ipld": "^0.30.2",
-    "ipld-in-memory": "^8.0.0",
     "standard": "^16.0.4",
     "swarmhash": "^0.1.1"
   },
   "scripts": {
     "lint": "standard update",
-    "update": "./update --max-files-per-batch 1",
+    "update": "./update.js --max-files-per-batch 1",
```
r0qs:
```suggestion
    "update": "node update.mjs --max-files-per-batch 1",
```

update.js
```diff
@@ -1,30 +1,30 @@
 #!/usr/bin/env node
```
r0qs: Since you renamed the file to add the `.js` extension we could use the `.mjs` instead. Also, you can remove the shebang line and change the `package.json` script to use node instead (see other comment).

update.js
```diff
@@ -1,30 +1,30 @@
...
-  if (done) {
-    throw new Error('Failed to calculate an IPFS hash.')
-  }
-
-  await iterator.return()
-  return value.cid.toString()
-}
+//this is a workaround for the semver as far as it is a CommonJS module
+import semver from "semver"
+const { compare, gt } = semver
```
r0qs: Not sure if I follow what the workaround is here. Why not do like the following?
```javascript
import { compare, gt } from "semver";
```

update.js
```diff
@@ -57,29 +57,29 @@ function updateSymlinkSync (linkPathRelativeToRoot, targetRelativeToLink) {
...
   if (targetRelativeToLink !== linkString) {
-    fs.symlinkSync(targetRelativeToLink, absoluteLinkPath, 'file')
+    symlinkSync(targetRelativeToLink, absoluteLinkPath, 'file')
     console.log('Created link ' + linkPathRelativeToRoot + ' -> ' + targetRelativeToLink)
   }
 }
 
 function updateCopy (srcRelativeToRoot, destRelativeToRoot) {
-  fs.readFile(path.join(__dirname, srcRelativeToRoot), function (err, data) {
+  readFile(join(__dirname, srcRelativeToRoot), function (err, data) {
```
r0qs:
```suggestion
  readFileSync(join(__dirname, srcRelativeToRoot), function (err, data) {
```

update.js
```diff
@@ -1,30 +1,30 @@
...
+  symlinkSync,
+  readFile,
+  readFileSync,
+  writeFile,
+  readdir,
+  stat,
+  lstat
+ } from "fs"
+import swarmhash from "swarmhash";
+import { promisify } from "util"
```
r0qs:
```suggestion
import { readFile } from 'node:fs/promises';
```
And remove `readFile` from the import above. Also, we could rename it to `readFileAsync` to be explicit about the function implementation being used:

```javascript
import { readFile as readFileAsync } from 'node:fs/promises';
```

update.js
```diff
@@ -141,11 +141,11 @@ async function makeEntry (dir, parsedFileName, oldList) {
   }
 
   if (!build.sha256 || !build.keccak256 || !build.urls || build.urls.length !== 2) {
-    const readFile = util.promisify(fs.readFile)
-    const fileContent = await readFile(absolutePath)
-    build.keccak256 = '0x' + ethUtil.keccak(fileContent).toString('hex')
+    const readfile = promisify(readFile)
+    const fileContent = await readfile(absolutePath)
```
r0qs: Since you are explicitly importing `readFileSync`, you can use [fs.Promises](https://nodejs.org/docs/latest-v16.x/api/fs.html#fspromisesreadfilepath-options) in places where the async version is desirable.

E.g.
```javascript
// assuming the following import:
import { readFile as readFileAsync } from 'node:fs/promises';
// ...
// you can do:
const fileContent = await readFileAsync(absolutePath);
```

```suggestion
    const fileContent = await readFileAsync(absolutePath);
```

r0qs (changes requested):
Hi @GeorgePlotnikov, thank you very much for the updates! I have some comments though. Let me know what you think ;)

package.json
```diff
@@ -1,22 +1,22 @@
...
+  "type": "module",
   "version": "0.0.0",
   "description": "Current and historical (emscripten) binaries for Solidity",
   "dependencies": {
     "semver": "^7.3.5"
   },
   "devDependencies": {
     "ethereumjs-util": "^7.1.3",
+    "blockstore-core": "^2.0.2",
     "ipfs-unixfs-importer": "^9.0.6",
```
r0qs: Please, also update `ipfs-unixfs-importer` to the latest version.
```suggestion
    "ipfs-unixfs-importer": "^11.0.1",
```

ipld-utils.js
0xGeorgii: fair enough, I thought to make it a bit easier but seems it is a redundant change

update.js
0xGeorgii: because `The requested module 'semver' is a CommonJS module, which may not support all module.exports as named exports.` I added it into the comment

update.js
0xGeorgii: thank you! good point!

update.js
0xGeorgii: I left it as `import semver from "semver"` with the further usage via `semver.compare` and `semver.gt`

r0qs: Hum, ok then. I thought `semver` supported both types.

r0qs (changes requested):
LGTM, but could you address this comment [here](https://github.com/ethereum/solc-bin/pull/125#discussion_r1018409092) as well?

0xGeorgii:
> LGTM, but could you address this comment [here](https://github.com/ethereum/solc-bin/pull/125#discussion_r1018409092) as well?

@r0qs thanks a lot for the review. Sorry it seems I missed it. Fixed.

package.json
```diff
@@ -1,22 +1,22 @@
...
-    "ipld-in-memory": "^8.0.0",
+    "blockstore-core": "^2.0.2",
+    "ipfs-unixfs-importer": "^11.0.1",
     "standard": "^16.0.4",
     "swarmhash": "^0.1.1"
   },
   "scripts": {
     "lint": "standard update",
-    "update": "./update --max-files-per-batch 1",
+    "update": "node ./update.mjs --max-files-per-batch 1",
```
r0qs: You can remove the `./` ;)
```suggestion
    "update": "node update.mjs --max-files-per-batch 1",
```

r0qs (changes requested):

r0qs (approved):
Thank you very much for your contribution @GeorgePlotnikov. I will just request the review of another dev and if everything is fine we will merge it.

0xGeorgii:
@r0qs thank you very much for your assistance!

package.json
```diff
@@ -1,22 +1,22 @@
...
-    "ipld-in-memory": "^8.0.0",
+    "blockstore-core": "^2.0.2",
+    "ipfs-unixfs-importer": "^11.0.1",
     "standard": "^16.0.4",
     "swarmhash": "^0.1.1"
   },
   "scripts": {
     "lint": "standard update",
-    "update": "./update --max-files-per-batch 1",
+    "update": "node update.mjs --max-files-per-batch 1",
```
cameel, line 19: Our release checklist in the main repo [refers to the `update` script by name](https://github.com/ethereum/solidity/blob/bf1a2c7cee3444480710db15123baca373af927b/ReleaseChecklist.md#L48). Can you submit a small PR there to replace it with `npm run update`?

cameel (approved):
Thanks. Clean and concise solution!

cameel:
For some reason the CircleCI lint job is not running here. We can ignore the bytecode comparison jobs even if they fail since they don't touch the `update` script but that one I'd like to see pass before we merge the PR.

cameel:
I forced it to run by creating a branch in the repo.

Also, I squashed the commits and restored the file comment (that seems to have disappeared in the changes).

package.json
0xGeorgii: @cameel [here](https://github.com/ethereum/solidity/pull/13743) it is please

r0qs:
@GeorgePlotnikov I took the freedom to fix the linter errors ;)

0xGeorgii:
> 

I am begging you 😆

cameel:
@r0qs You need to move the branch on our side to the latest commit to have CI run here. It does not seem to run on the fork.

cameel:
@r0qs Review fixes should be squashed and then feel free to merge.

PR #131, opened 2023-02-22
Fix nodejs version to 16 in nightly builds

r0qs:
The docker runner images of Github actions had the default nodejs version updated to `18`, see: https://github.com/actions/runner-images/issues/7002

Sadly, some CI jobs, like the [nightly builds](https://github.com/ethereum/solc-bin/actions/runs/4238005281/jobs/7365127264) started to fail due to requirement of nodejs version`16`, needed for the `sha3` library: https://github.com/axic/keccakjs/issues/13. So we are fixing the nodejs version for now until we fix the issue.

### DIFF
--- a/.github/workflows/nightly-emscripten.yml
+++ b/.github/workflows/nightly-emscripten.yml
@@ -118,6 +118,11 @@ jobs:
 
     if: "needs.build-emscripten-nightly.outputs.nightly-already-exists == 'false'"
     steps:
+      # TODO: remove when https://github.com/axic/keccakjs/issues/13 got fixed
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+
       - name: Remove unused software to free up space
         run: |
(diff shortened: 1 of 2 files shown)


cameel:
@r0qs Is it the same thing we ended up merging in the PR with binaries? Or a different case?

r0qs:
You mean if it is the same problem? No, this is specifically the problem of using https://github.com/axic/keccakjs/issues/13 with a recent version of nodejs. More specifically, it is the exact error on the description of this PR: https://github.com/ethereum/solc-bin/pull/101#issue-1045827371

See here our current GH action logs: https://github.com/ethereum/solc-bin/actions/runs/4238005281/jobs/7365127264

But it is related with the other problem in the sense that it was triggered by the same event, i.e. the update of the default nodejs version from 16 to 18 in the docker images of Github actions runners.

cameel (approved):

PR #142, opened 2024-03-13
Free up disk space in GitHub actions, round 2

cameel:
CI jobs are running out of space so we need another round of #86.

~A draft for now because first I need to figure out what can be deleted.~ Actually, this time I managed to solve the issue without deleting any files.

### DIFF
--- a/.github/workflows/nightly-emscripten.yml
+++ b/.github/workflows/nightly-emscripten.yml
@@ -20,13 +20,14 @@ jobs:
       nightly-already-exists: ${{ env.NIGHTLY_ALREADY_EXISTS }}
 
     steps:
-      - name: Remove unused software to free up space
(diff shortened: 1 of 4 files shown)


cameel:
We're in luck. Looks like there's another partition mounted at `/mnt` with enough space so we can drop the file removal hack.
```
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        73G   53G   21G  73% /
tmpfs           7.9G  172K  7.9G   1% /dev/shm
tmpfs           3.2G  1.1M  3.2G   1% /run
tmpfs           5.0M     0  5.0M   0% /run/lock
/dev/sda15      105M  6.1M   99M   6% /boot/efi
/dev/sdb1        74G  4.1G   66G   6% /mnt
tmpfs           1.6G   12K  1.6G   1% /run/user/1001
```

PR #145, opened 2024-03-14
Fix macosx-amd64/solc-macosx-amd64-v0.8.25+commit.b61c2a91.

aarlt:
Recreate macosx-amd64/solc-macosx-amd64-v0.8.25+commit.b61c2a91 using ad-hoc signature.


r0qs:
Don't we also need to adjust the hashes? https://github.com/ethereum/solc-bin/blob/a0a5e364001694c281576dae3bd6854b6d6b7782/macosx-amd64/list.json#L1100

r0qs (approved):

PR #147, opened 2024-04-30
Drop swarmhash support

### LINKED ISSUE #14700: Incompatibility between swarmhash and node.js > 16 makes releases difficult
solc-bin depends on [swarmhash](https://github.com/axic/swarmhash), which depends on [keccakjs](https://github.com/axic/keccakjs), which depends on [sha3](https://www.npmjs.com/package/sha3). [sha3 fails to build on node.js > 16](https://github.com/axic/keccakjs/issues/13). Now that [node.js 16 has reached EOL](https://nodejs.org/en/blog/announcements/nodejs16-eol), distros like Arch Linux are dropping that version and it starts being more and more of a hindrance in our release process. Recalculating hashes locally requires workarounds like running the `update` script in a Docker image.

I reported the issues in keccakjs a long time ago (https://github.com/axic/keccakjs/issues/13) but they were not fixed so far. There's also a [PR to switch swarmhash](https://github.com/axic/swarmhash/pull/5) to the actively maintained [keccak](https://www.npmjs.com/package/keccak) package, which would be an even better way to solve the issue. I pinged @axic in the PR, and there's a chance it will get merged. If not, we'll have to look for an alternative. I'm preemptively creating an issue for it here so that we can track it on the focus board.

Maybe we should also consider dropping the swarm hashes altogether from the file list. AFAIK solc-bin is not hosted on Swarm anyway and the project has gone through significant changes over the years, which may have even invalidated those hashes. We should check if having these hashes there still makes sense.

r0qs:
This PR removes support for swarm hashes from the binary releases list and updates the Node.js dependency to its latest version in the CI.

Additionally, the PR updates the GitHub Actions to utilize their latest versions. Some of them, such as https://github.com/actions/download-artifact and https://github.com/actions/upload-artifact, were utilizing deprecated versions. Only actions maintained by Github were updated.

```
actions/upload-artifact@v3 is scheduled for deprecation on November 30, 2024. [Learn more.](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/)
Similarly, v1/v2 are scheduled for deprecation on June 30, 2024. Please update your workflow to use v4 of the artifact actions.
This deprecation will not impact any existing versions of GitHub Enterprise Server being used by customers.
```

Fixes https://github.com/ethereum/solidity/issues/14700

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,8 +3,7 @@ version: 2.1
 jobs:
   chk_npm_lint_and_test:
     docker:
-      # FIXME: sha3 package fails to build on latest node.js (version 17)
-      - image: cimg/node:16.20
+      - image: cimg/node:current
     resource_class: small
     steps:
       - checkout
--- a/.github/workflows/nightly-emscripten.yml
+++ b/.github/workflows/nightly-emscripten.yml
@@ -119,10 +119,7 @@ jobs:
 
     if: "needs.build-emscripten-nightly.outputs.nightly-already-exists == 'false'"
     steps:
-      # TODO: remove when https://github.com/axic/keccakjs/issues/13 got fixed
-      - uses: actions/setup-node@v3
-        with:
-          node-version: 16
+      - uses: actions/setup-node@v4
 
       - name: Symlink solc-bin to /mnt/solc-bin
         # It's now too big to fit on the main partition on Ubuntu, which is mostly filled with software.
--- a/.github/workflows/nightly-emscripten.yml
+++ b/.github/workflows/nightly-emscripten.yml
@@ -29,7 +29,7 @@ jobs:
           sudo chown "$USER" /mnt/solc-bin/
           ln -s /mnt/solc-bin/ solc-bin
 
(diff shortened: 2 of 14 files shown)


OKEAMAH (approved):
Hjj

OKEAMAH (commented):
Nn

cameel (approved):
Let's do it.

Like I said on the call, the fact that the position of the IPFS hash changes is a potential problem. But these hashes are probably not that widely used anyway, so I think we can still do it without affecting most tools. We could insert `null` in place of Swarm's hash, but tools should recognize these URLs based on the protocol and not rely on the position anyway.

Please remember to update the docs though.

cameel:
By the way, the PR probably needs an update before we merge it, to account for nightlies that were added in the meantime. I rebased it, but now it will probably fail because of those nightlies.

cameel:
Note that if you come back to this PR tomorrow or later, you'll need to remember to rebase again before merging since we'll have a new nightly by then :)

PR #151, opened 2024-05-13
Reduce polling frequency of turnstyle GH action to avoid API rate limit for unauthenticated requests

r0qs:
We are performing unauthenticated request to GitHub API when running the job to mirror the binaries to S3 which [imposes a limited number of API request](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-unauthenticated-users). See: https://github.com/ethereum/solc-bin/actions/runs/9055823647/job/24877483141#step:3:9

Passing the GH token (or a token with `actions:read` permission) for authentication should increase our rate limit of requests per hour. Alternatively,  we can increase the waiting time between checks for previous workflows completion done by the [turnstyle](https://github.com/softprops/turnstyle?tab=readme-ov-file#inputs) action.

### DIFF
--- a/.github/workflows/s3-mirror.yml
+++ b/.github/workflows/s3-mirror.yml
@@ -33,6 +33,7 @@ jobs:
         uses: softprops/turnstyle@v1
         with:
           same-branch-only: no
+          poll-interval-seconds: 120 # in seconds
 
       - name: Configure the S3 client
         run: |


.github/workflows/s3-mirror.yml
```diff
@@ -31,6 +31,8 @@ jobs:
       - name: Wait for other instances of this workflow to finish
         # It's not safe to run two S3 sync operations concurrently with different files
         uses: softprops/turnstyle@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           same-branch-only: no
```
cameel: Not great that we have to give it the token. It's a third-party action so always more risk of it being exfiltrated. Still, this is probably a bit too paranoid, we'll probably be ok.

Still, odd that this is doing enough API requests to hit the limits. I'd only expect that with lots of stuff being merged at the same time. So possible that we're only hitting the limit here because we used it up elsewhere.

cameel (approved):

.github/workflows/s3-mirror.yml
r0qs: Should be resolved by reducing the polling frequency of [turnstyle](https://github.com/softprops/turnstyle) github action, the default is every minute: https://github.com/softprops/turnstyle?tab=readme-ov-file#inputs

cameel (approved):

PR #153, opened 2024-05-21
Binaries for 0.8.26

r0qs:


bin/list.js
```diff
@@ -1,5 +1,6 @@
 
 var soljsonSources = [
+  "soljson-v0.8.26+commit.8a97fa7a.js",
```
nikola-matic, line 3: This is weird, shouldn't it be a nightly like the previous ones?

nikola-matic: Ah never mind, it was like this in the previous binary pushes.

nikola-matic (approved):

bin/list.js
r0qs: I got a bit confused as well, but I looked at the previous PRs and they also update it. See: https://github.com/ethereum/solc-bin/pull/140/files and https://github.com/ethereum/solc-bin/pull/144/files

PR #154, opened 2024-08-21
Update CI to checkout solidity submodules

### LINKED ISSUE #15350: Recursively clone solidity repository on solc-bin nightly builds.
Now that https://github.com/ethereum/solidity/pull/15195 was merged we require to recursively clone Solidity repository in solc-bin when running the github actions for nightly builds as well as bytecode compare jobs. See: https://github.com/ethereum/solc-bin/actions/runs/10481074451/job/29029908615#step:7:496

This seems to be trivially fixed by adding `submodules: 'recursive'` option to the [actions/checkout](https://github.com/actions/checkout) in the CI.

See successful test here: https://github.com/ethereum/solc-bin/actions/runs/10487749944/job/29048779836

r0qs:
Fixes https://github.com/ethereum/solidity/issues/15350

### DIFF
--- a/.github/workflows/random-macosx-build.yml
+++ b/.github/workflows/random-macosx-build.yml
@@ -69,6 +69,7 @@ jobs:
           ref: v${{ env.SOLIDITY_VERSION }}
           path: solidity/
           fetch-depth: 0
+          submodules: 'recursive'
 
       - name: Install semver utility
         run: |
--- a/.github/workflows/t-bytecode-compare.yml
+++ b/.github/workflows/t-bytecode-compare.yml
@@ -51,6 +51,7 @@ jobs:
           path: 'solidity/'
           # bytecode_reports_for_modified_binaries.sh requires access to a working copy with full history
           fetch-depth: 0
+          submodules: 'recursive'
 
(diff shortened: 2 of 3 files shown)


.github/workflows/random-macosx-build.yml
```diff
@@ -69,6 +69,7 @@ jobs:
           ref: v${{ env.SOLIDITY_VERSION }}
           path: solidity/
           fetch-depth: 0
+          submodules: 'recursive'
```
r0qs, line 72: This action seems to be deactivated, but I added the option here anyway for sanity.

.github/workflows/t-bytecode-compare.yml
```diff
@@ -88,6 +89,7 @@ jobs:
         with:
           repository: 'ethereum/solidity'
           path: 'solidity/'
+          submodules: 'recursive'
```
r0qs, line 92: This may not be required here either. No build is done here.

clonker (approved):
Don't think it hurts to have a couple more `recursive` flags than absolutely necessary.

PR #155, opened 2024-09-04
Release binaries for v0.8.27

clonker:


r0qs (approved):

.github/workflows/t-bytecode-compare.yml
```diff
@@ -29,6 +29,30 @@ jobs:
...
 
+      - name: Free Disk Space Before Build
+        if: "env.PLATFORM == 'macosx-amd64'"
+        run: |
+          df -hI /dev/disk3s1s1
+          sudo rm -rf /Applications/Xcode_14.3.1.app
+          sudo rm -rf /Applications/Xcode_15.0.1.app
+          sudo rm -rf /Applications/Xcode_15.1.app
+          sudo rm -rf /Applications/Xcode_15.2.app
+          sudo rm -rf /Applications/Xcode_15.3.app
```
r0qs, line 40: Is this always installed? I'm wondering if this will not break eventually. I.e. when the base `macos-latest` image stop to provide them (currently it is macos-14): https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md

r0qs: Maybe we need to create an issue to rethink the cleanup process before the next release.

r0qs: Oh, you used `--force`, so it will ignore nonexistent files. So yeah, it will not break.

clonker: IMO this shouldn't be more than a temporary fix in any case.

cameel: > So yeah, it will not break.

Well, it *will* break, because when it stops deleting stuff, we'll run out of space again :)

Not sure if we can do better though. The macOS runner does not have the extra data partition so to fit the repo in we do have to delete stuff. We should just make this a bit more robust by not depending on specific versions of stuff we delete.

r0qs: Yeah, what I mean by breaking was related to the `rm` command failing to find the files. But yeah, it will run out of space again for sure. Maybe we could have a script that check for the current disk space and erase old dated files. Not the best approach either, but maybe a bit better than what we currently do.

cameel: An alternative would be to go back to deleting old nightlies. We have so many now that [github crashes when you go to `bin/`](https://github.com/ethereum/solc-bin/tree/gh-pages/bin). It's just that it also does not feel like a proper solution because we can't delete them from history so we're still storing them there. The repo will still keep growing in an unbounded way, only the checkout will be smaller.

I don't think it was ever a good idea to store nightly binaries in a github repo (they really should have been stored as something less permanent, e.g. pre-releases, if at all).

clonker: It's still better than not deleting it b/c you can do a shallow clone (theoretically at least if we ignore the bytecode compare). Is anyone relying on the nightlies being here? Otherwise we could just not store them here anymore and people can fetch them from CI as artifacts?

cameel: You can access them e.g. in Remix, so some people may be using them. Some frameworks may be making them available to users too - I remember e.g. that there was some interest from Truffle in having them accessible in the new directory structure. But I don't think anyone is relying on them in production. And they do, we don't have to support that (I wonder if e.g. Sourcify does?).

I added this for discussion next Wednesday. Perhaps it's time to drop some historical baggage and simplify things. I'd rather avoid half-measures, but I'm also not sure how far we can safely go.

TBH I wouldn't be against just scrapping those nightlies completely. They're a nuisance while at the same time they are not useful enough. It's often much more convenient to get a platform-specific binary from `develop` and we don't have those here. And also it's often more useful for us to share a build from an unmerged PR. Both of these things we currently get from CircleCI and we don't even notice when the nightlies here break. CircleCI only stores artifacts for 30 days, but that's usually more than enough. The only annoying bit is that giving them to people always requires explanations, because links are a buried deep in the CCI UI.

PR #156, opened 2024-09-11
Use symlink for bin/soljson-latest.js

### LINKED ISSUE #13508: Replace `bin/soljson-latest.js` in solc-bin with a symlink
We tried to make it a symlink in https://github.com/ethereum/solc-bin/pull/33 but back then we still served binaries via Github Pages and that does not support symlinks so we reverted the change in https://github.com/ethereum/solc-bin/pull/44. Now that we switched to AWS for hosting, we can revisit the idea.

clonker:
Fixes https://github.com/ethereum/solidity/issues/13508

Related PR: https://github.com/ethereum/solc-bin/pull/33

### DIFF
--- a/update.mjs
+++ b/update.mjs
@@ -11,7 +11,6 @@ import {
   readFileSync,
   writeFile,
   readdir,
-  stat,
   lstat
 } from 'fs'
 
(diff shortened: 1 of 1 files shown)


cameel (approved):
Any idea why it wasn't working with a copy though? Was it running into an error but silencing it or something? If that's the case we should fix that, to avoid this kind of breakage in the future.

Also, have you checked that the link does get properly updated now when you add a new release binary (locally)? I'd assume yes since now it goes through the same path that other dirs go, but still, I don't see why the other path wasn't working in the first place so better to double check. If it does not work, we likely won't notice during the next release.

clonker:
I didn't investigate much why it broke in the first place since that functionality was only used for that one place and this PR removes it - not fixing it here of course still carries the danger that if we need the functionality back for some reason and someone remembers this but doesn't remember it was not working correctly and just takes it from the history without testing it again...then... :D
A silenced error seems likely, yes.

And yeah, I checked locally if the link gets created/updated. Works on my machine! And running it takes forever, we should really get rid of (at least some!) nightly builds... Can't imagine anyone is using nightlies back from 2020.

clonker:
@cameel you made me curious what may be the issue and (without having confirmed it really) it looks like `readFileSync` is the culprit. It should have been `readFile`, probably, the `Sync` version [does not accept the callback function](https://www.geeksforgeeks.org/node-js-fs-readfilesync-method/) :)

PR #162, opened 2025-09-11
Update org name

r0qs:
There is a section on the [README](https://github.com/argotorg/solc-bin?tab=readme-ov-file#deprecation-notice-for-the-ethereumgithubio-domain) that I didn't update because I think we should maybe remove it completely, or at least change it to  `argotorg.github.io` since now we have control over the domain. Open for suggestions.

### DIFF
--- a/.github/workflows/nightly-emscripten.yml
+++ b/.github/workflows/nightly-emscripten.yml
@@ -8,7 +8,7 @@ on:
 env:
   TARGET_BRANCH: gh-pages
   COMMITTER_NAME: emscripten nightly action
-  COMMITTER_EMAIL: builds@ethereum.org
+  COMMITTER_EMAIL: builds@argot.org
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
@@ -31,7 +31,7 @@ jobs:
 
       - uses: actions/checkout@v4
         with:
-          repository: 'ethereum/solidity'
+          repository: 'argotorg/solidity'
           ref: 'develop'
           path: 'solidity/'
           submodules: 'recursive'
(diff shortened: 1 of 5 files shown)


.github/workflows/nightly-emscripten.yml
```diff
@@ -8,7 +8,7 @@ on:
 env:
   TARGET_BRANCH: gh-pages
   COMMITTER_NAME: emscripten nightly action
-  COMMITTER_EMAIL: builds@ethereum.org
+  COMMITTER_EMAIL: builds@argot.org
```
r0qs, line 11: Does this email need to exist, or is it only used for commit messages?

cameel:
> There is a section on the [README](https://github.com/argotorg/solc-bin?tab=readme-ov-file#deprecation-notice-for-the-ethereumgithubio-domain) that I didn't update because I think we should maybe remove it completely, or at least change it to  `argotorg.github.io` since now we have control over the domain. Open for suggestions.

You mean the bit about https://ethereum.github.io/solc-bin/? That broke with the transition to `argotorg`, right? Or was it broken already for some time?

In any case, since it does not seem like anyone complained, so we can remove it. No point switching to `argotorg.github.io` because it was only there to keep it working for anyone who might have been using it back when we added `binaries.soliditylang.org`. If they can switch, they can just as well use the latter.

cameel:
By the way, this means that the redirect github creates when a repo is moved does not work for Github Pages. Good that we did not have anything important there. This is the kind of hidden breakage I was afraid of with the switch. There are always things like this that you only realize once you do it.

.github/workflows/nightly-emscripten.yml
cameel: Only for commit messages. See https://github.com/argotorg/solc-bin/pull/30#discussion_r450041505.

cameel (approved):
Aside from removal of the outdated bit from README, looks ok.

r0qs:
> You mean the bit about https://ethereum.github.io/solc-bin/? That broke with the transition to `argotorg`, right? Or was it broken already for some time?

Yeah, it broke with the migration.

cameel (approved):

PR #169, opened 2025-11-04
Update sync script to support Cloudflare R2 and CDN cache purging

marcocastignoli:
### Instructions to test the script

The script will delete the .git folder, so before proceeding

```bash
tmpdir=$(mktemp -d) && cd "$tmpdir"
cp /Users/marcocastignoli/Projects/solc-bin/sync-r2.sh .
git init
echo "hello r2" > nightly-build.txt
git add sync-r2.sh nightly-build.txt
git commit -m "Test payload for R2 sync"
```

Then set the environment variables

```bash
export R2_ACCOUNT_ID=
export R2_ZONE=
export R2_BUCKET_NAME=
export CLOUDFLARE_ZONE_ID=
export CLOUDFLARE_CACHE_HOST=
export CLOUDFLARE_API_TOKEN=
export AWS_ACCESS_KEY_ID=
export AWS_SECRET_ACCESS_KEY=
export AWS_DEFAULT_REGION=
```

Finally run the script

```bash
./sync-r2.sh "$R2_ACCOUNT_ID" "$R2_ZONE" "$R2_BUCKET_NAME" "$CLOUDFLARE_ZONE_ID" "$CLOUDFLARE_CACHE_HOST" "$CLOUDFLARE_API_TOKEN"
```


sync-s3.sh
```diff
@@ -55,22 +77,39 @@ find . \
     -regex "^\(.*/\)*[^/]*\+[^/]*$" \
     -exec bash -c 'ln --symbolic --no-target-directory "$(basename "{}")" "$(dirname "{}")/$(basename "{}" | tr "+" " ")"' \;
 
-echo "===> Syncing binaries with the S3 bucket"
```
d-xo, line 58: I think we should probably keep at least the s3 sync in place for now (and maybe the cloudfront invalidation too?) just as a backup for a bit in case something doesn't work as expected with cloudfront?

marcocastignoli: I added another script specifically for r2 and updated the `.github/workflows/s3-mirror.yml` file to call the new script

marcocastignoli:
I tested the script locally with some dummy files and it's working. Soon @clonker will run the script with a local copy he has of the full repository.

sync-s3.sh
```diff
@@ -73,4 +73,4 @@ aws cloudfront create-invalidation \
         /soljson.js \
         $(find . -wholename '*/list.*' | cut --characters 2-) \
         $(find . -wholename '*/*-latest' | cut --characters 2-) \
-        $(find . -wholename '*/*-latest.*' | cut --characters 2-)
+        $(find . -wholename '*/*-latest.*' | cut --characters 2-)
```
nikola-matic: Add back the newline pls

sync-r2.sh
```diff
@@ -0,0 +1,105 @@
...
+)
+while IFS= read -r path; do
+    purge_paths+=("/${path}")
+done < <(find . -wholename '*/list.*' | cut --characters 2-)
+while IFS= read -r path; do
+    purge_paths+=("/${path}")
+done < <(find . -wholename '*/*-latest' | cut --characters 2-)
+while IFS= read -r path; do
+    purge_paths+=("/${path}")
+done < <(find . -wholename '*/*-latest.*' | cut --characters 2-)
```
r0qs: Maybe we could rewrite that like below? wdyt?
```suggestion
while IFS= read -r path; do
      purge_paths+=("/${path}")
done < <(find . \( -wholename '*/list.*' -o -wholename '*/*-latest' -o -wholename '*/*-latest.*' \) | cut --characters 2-)
```

r0qs (commented):
Nice. Thanks a lot for that @marcocastignoli! It looks good to me. Just a small comment.

sync-r2.sh
```diff
@@ -0,0 +1,105 @@
...
+
+purge_payload='{"files":['
+separator=""
+for path in "${purge_paths[@]}"; do
+    purge_payload+="${separator}\"https://${cloudflare_cache_host}${path}\""
+    separator=","
```
nikola-matic: I assume the trailing comma at the end of `purge_payload` will not cause problems?

sync-r2.sh
```diff
@@ -0,0 +1,105 @@
...
+
+echo "===> Syncing binaries with the Cloudflare R2 bucket"
+aws --endpoint-url "$r2_endpoint" s3 sync . "$r2_bucket_uri" --delete --follow-symlinks --no-progress --exclude "*/list.*"
+
+echo "===> Syncing file lists with the Cloudflare R2 bucket"
+aws --endpoint-url "$r2_endpoint" s3 sync . "$r2_bucket_uri" --delete --follow-symlinks --no-progress --exclude "*" --include "*/list.*"
+
+echo "===> Purging Cloudflare CDN cache"
+# Purge only the files that might change in-place when new binaries are added.
+purge_paths=(
```
d-xo, line 75: As we discussed a bit in chat: since bandwidth in cloudflare is free, and read requests are very cheap, is there any reason to not just invalidate the entire cache each time we upload for the sake of simplicity?

marcocastignoli: It actually makes sense, I forgot about our discussion yesterday ahah

if also @r0qs agrees we can just remove the whole Cloudflare cache deletion

r0qs: But why clean the cache of previous binaries? I mean, for new binaries, we would only need to clear the cache of the metadata (i.e. lists) files and "latest" links I think. No?

marcocastignoli: The idea is that with Cloudflare everything is so cheap that we will not even have to clean specific files, but we can invalidate everything

r0qs: Hum... but this would invalidate the entire cache every day when there's a nightly build (https://github.com/argotorg/solc-bin/blob/8b8bc6f3988f45e161d757d7af432272865dd032/.github/workflows/nightly-emscripten.yml), which somewhat defeats the purpose of having a cache.

sync-r2.sh
marcocastignoli: it's not at the end, it's at the beginning. the separator is "" for the first one (prefix) and then `, "others"`

sync-r2.sh
marcocastignoli: @d-xo actually I think it makes sense to keep the clear cache mechanism. I tested it and we are successfully calling the Cloudflare API

d-xo: fine by me

.github/workflows/mirror-content.yml
cameel, line 1: Script name, description and job name need to be updated. It no longer handles just Amazon's S3, but they all still refer to it.

Maybe let's call it `mirror-content.yml`?

sync-r2.sh
```diff
@@ -0,0 +1,99 @@
...
+find . -path './.*' -delete
+find . -path './_*' -delete
+
+echo "===> Adding compatibility symlinks for files containing plus signs in the name"
+# NOTE: This is a quick'n'dirty workaround for Amazon S3 decoding plus sign in paths
+# as a space even though this substitution is only supposed to happen in a query string.
+# See https://forums.aws.amazon.com/thread.jspa?threadID=55746
+find . \
+    -regex "^\(.*/\)*[^/]*\+[^/]*$" \
+    -exec bash -c 'ln --symbolic --no-target-directory "$(basename "{}")" "$(dirname "{}")/$(basename "{}" | tr "+" " ")"' \;
```
cameel: I wonder if R2 actually needs this workaround. S3 did when I wrote the script, but it's possible that Cloudflare did not copy the API bug for bug :)

All binaries have `+` in their names, so being able to drop it would halve our storage use and make the sync faster. So perhaps worth checking. Though in the end the thing we pay the most for is bandwidth so it won't affect cost much.

If you decide to leave it as is, without checking, I'd at least add a comment that we have not checked and are not sure if it's still needed for Cloudflare.

sync-r2.sh
```diff
@@ -0,0 +1,99 @@
...
+    separator=","
+done
+purge_payload+=']}'
+
+curl --fail --show-error --silent \
+    -X POST "https://api.cloudflare.com/client/v4/zones/${cloudflare_zone_id}/purge_cache" \
+    -H "Authorization: Bearer ${cloudflare_api_token}" \
+    -H "Content-Type: application/json" \
+    --data "$purge_payload" \
+    >/dev/null
```
cameel: Not sure what the response looks like here, so maybe there is a good reason to hide it (very long, contains secrets, etc), but as a general rule I'd leave things like this in the output. Sometimes it helps figure out what's wrong when things do not work as expected.

```suggestion
curl --fail --show-error --silent \
    -X POST "https://api.cloudflare.com/client/v4/zones/${cloudflare_zone_id}/purge_cache" \
    -H "Authorization: Bearer ${cloudflare_api_token}" \
    -H "Content-Type: application/json" \
    --data "$purge_payload"
```

sync-r2.sh
```diff
@@ -0,0 +1,99 @@
...
+    purge_paths+=("/${path}")
+done < <(find . \( -wholename '*/list.*' -o -wholename '*/*-latest' -o -wholename '*/*-latest.*' \) | cut --characters 2-)
+
+purge_payload='{"files":['
+separator=""
+for path in "${purge_paths[@]}"; do
+    purge_payload+="${separator}\"https://${cloudflare_cache_host}${path}\""
+    separator=","
+done
+purge_payload+=']}'
```
cameel: I see you're going out of your way to handle whitespace properly here (the original script just assumed we won't ever use that in file names), but then this will still not handle some paths properly (anything containing quotes or backslashes). If we assume `jq` is installed, we can easily handle those too by having it escape the values for us while also making it simpler:

```suggestion
jq --null-input '{"files": $ARGS.positional}' --args -- "${purge_paths[@]}"
```

cameel (commented):
Job name should be updated and perhaps the PR could use a few tweaks (see my comments), but overall seems good.

sync-s3.sh
cameel: Just a heads up that if you run both scripts, then they will process the files twice. The second run repeats the updating of timestamps, removal of hidden files and creation of symlinks that the S3 version has done already. Fortunately all of these actions are idempotent, so this should not break anything and at most just do some unnecessary work, but the idempotence becomes an implicit assumption we should now document. Can you add a comment about this? It could bite us in the future if we try to add something there that does not satisfy this property.

sync-r2.sh
```diff
@@ -0,0 +1,92 @@
...
+    "/soljson.js"
+)
+while IFS= read -r path; do
+    purge_paths+=("/${path}")
+done < <(find . \( -wholename '*/list.*' -o -wholename '*/*-latest' -o -wholename '*/*-latest.*' \) | cut --characters 2-)
+
+purge_payload="$(jq --null-input \
+    --arg host "https://${cloudflare_cache_host}" \
+    '{"files": ($ARGS.positional | map($host + .))}' \
+    --args -- "${purge_paths[@]}")"
```
cameel:
```suggestion
    --args -- "${purge_paths[@]}"
)"
```

cameel (approved):
Looks good to me. Let's try it out!

cameel:
Oh, just please squash the commits before merging.

Does not have to be one commit, but if there are more, each one should be one logical change that goes straight to the point, without all the back and forth from the review comments, because that's just noise when you look at history and try to figure out what changed.

PR #170, opened 2025-11-27
Create two separate actions for R2 and S3 upload

marcocastignoli:
With the previous PR (#169), the r2 mirror script can no longer run because the .git files are removed. In general, both scripts run in the same environment, causing the second script to repeat the timestamp updates, removal of hidden files, and creation of symlinks that the S3 version has already performed.

With this PR, we introduce two separate GitHub actions: one for R2 and one for S3.

### DIFF
--- a/.github/workflows/r2-mirror.yml
+++ b/.github/workflows/r2-mirror.yml
@@ -1,0 +1,65 @@
+name: Mirror repository content to R2 storage bucket
+
+on:
+  push:
+    branches:
+      - gh-pages
+
+  schedule:
+    # Pushes to `gh-pages` done from other actions cannot trigger this action so we also want it to run
+    # on a schedule. Let's give the nightly job a 1h head-start and run every day at 1:00.
+    - cron: "0 1 * * *"
+
+env:
+  R2_BUCKET: solc-bin
+  R2_ACCOUNT_ID: 1fbd136205d2c780838f0d2c014bf69c
+  R2_ZONE: EEUR
+  CLOUDFLARE_ZONE_ID: d63261269eeac3abf4c1b2f52e71ae1b
+  CLOUDFLARE_CACHE_HOST: binaries.soliditylang-test.org
+
+jobs:
+  mirror-content:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/s3-mirror.yml
```diff
@@ -0,0 +1,58 @@
...
+          sudo mkdir /mnt/solc-bin/
+          sudo chown "$USER" /mnt/solc-bin/
+          ln -s /mnt/solc-bin/ solc-bin
+
+      - name: Wait for other instances of this workflow to finish
+        # It's not safe to run two S3 sync operations concurrently with different files
+        uses: softprops/turnstyle@v1
+        with:
+          same-branch-only: no
+          poll-interval-seconds: 120 # in seconds
```
cameel, line 36: For a moment I was worried that this might prevent the R2 sync from running concurrently with the S3 one. But looking at the source, the action only waits if the workflow name matches, so it should be fine. I originally expected that you were going to put both jobs in the same workflow, but if you did that, it would block, so good that you did not.

Still, after we merge, we should check to see if both are indeed running at the same time.

.github/workflows/r2-mirror.yml
cameel, line 1: Would be good to also update the comment I asked you to add in https://github.com/argotorg/solc-bin/pull/169#discussion_r2560899808 to say that the actions are not in fact idempotent and that each script will only work on a fresh checkout.

cameel (approved):
Some comments need to be updated, but other than that the code seems fine, so feel free to merge whenever you're ready.

cameel:
As a side note, we should also pin those actions. Back when I wrote this we weren't doing that consistently, but now that's what we normally do. Looks like github even recently added an option to enforce that in PRs ([Enforce SHA pinning](https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/#enforce-sha-pinning)) and we should enable it. But that's just FYI. It's been like that before and it's not strictly related to the PR, just something I noticed while reviewing it. I'll add this as an org TODO.

PR #173, opened 2025-12-10
Update cloudflare cache host in R2 sync script

r0qs:
For some unknown reason, the checkout of solc-bin that was taking around 6 minutes (https://github.com/argotorg/solc-bin/actions/runs/20085596221/job/57622225915), now takes around 24 minutes (https://github.com/argotorg/solc-bin/actions/runs/20100028534/job/57668159518).

This is not expected, but maybe it is related to the cache host that was wrong and was pointing to test domain. I did a small test in another PR and the checkout was faster when using the correct domain, so I'm confident this will solve the issue.


marcocastignoli (approved):
Right! I completely forgot about this. Lgtm

PR #178, opened 2025-12-18
Fix `purge_paths` in `sync-r2.sh` by removing unnecessary leading slashes

marcocastignoli:
`purge_paths`, a list containing the files to be purged was created by adding resources with the wrong path "https://binaries.soliditylang.org//bin/list.json" instead of "https://binaries.soliditylang.org/bin/list.json" (notice the double slash before `bin/list.json`).

### DIFF
--- a/sync-r2.sh
+++ b/sync-r2.sh
@@ -75,13 +75,16 @@ purge_paths=(
 )
 while IFS= read -r path; do
     purge_paths+=("/${path}")
-done < <(find . \( -wholename '*/list.*' -o -wholename '*/*-latest' -o -wholename '*/*-latest.*' \) | cut --characters 2-)
+done < <(find . \( -wholename '*/list.*' -o -wholename '*/*-latest' -o -wholename '*/*-latest.*' \) | cut -c 2-)
 
 purge_payload="$(jq --null-input \
     --arg host "https://${cloudflare_cache_host}" \
     '{"files": ($ARGS.positional | map($host + .))}' \
     --args -- "${purge_paths[@]}"
 )"
+
+# Print the payload to check purged files
+echo $purge_payload
 
 curl --fail --show-error --silent \
     -X POST "https://api.cloudflare.com/client/v4/zones/${cloudflare_zone_id}/purge_cache" \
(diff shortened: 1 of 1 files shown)


sync-r2.sh
```diff
@@ -74,15 +74,18 @@ purge_paths=(
     "/soljson.js"
 )
 while IFS= read -r path; do
-    purge_paths+=("/${path}")
-done < <(find . \( -wholename '*/list.*' -o -wholename '*/*-latest' -o -wholename '*/*-latest.*' \) | cut --characters 2-)
+    purge_paths+=("${path}")
+done < <(find . \( -wholename '*/list.*' -o -wholename '*/*-latest' -o -wholename '*/*-latest.*' \) | cut -c 2-)
```
r0qs, line 78: Why the change to `-c` is needed? I see the commit message: "Correct cut command in sync-r2.sh" but I don't know for what reason you cannot use `--characters`. The man page says:
```txt
-c, --characters=LIST select only these characters
```

So both options should be valid. Is this an issue with the current version of `cut` being used by the GH action?

r0qs (commented):
Besides one comment. The PR looks good to me.

sync-r2.sh
marcocastignoli: no, sorry it's just that on osx (where I tested the script) cut doesn't support --character but only -c.

sync-r2.sh
```diff
@@ -74,15 +74,18 @@ purge_paths=(
...
 
 purge_payload="$(jq --null-input \
     --arg host "https://${cloudflare_cache_host}" \
     '{"files": ($ARGS.positional | map($host + .))}' \
     --args -- "${purge_paths[@]}"
 )"
 
+# Print the payload to check purged files
+echo $purge_payload
```
cameel:
```suggestion
echo "$purge_payload"
```

sync-r2.sh
cameel: That's why we can't have nice things :(

The outdated shell utils that macOS ships with are a pain.

r0qs: > no, sorry it's just that on osx (where I tested the script) cut doesn't support --character but only -c.

Right, so can you revert that change?

r0qs (approved):
I'm fine with the change in https://github.com/argotorg/solc-bin/pull/178/changes#r2636402198 if that makes easier to test locally in different environments.

PR #180, opened 2026-02-18
Release v0.8.34 binaries

rodiazet:


package-lock.json
nikola-matic, line 1: Remove

solc-bin-binaries.tar
nikola-matic, line 1: Remove

cameel:
For the record, in case we're later wondering why there are two PRs for this. This one was reverted and redone as #183.

PR #188, opened 2026-07-15
Fix unauthenticated rate limit errors in the R2 mirror workflow

marcocastignoli:
The scheduled r2-mirror runs keep failing at the turnstyle step with API rate limit exceeded for 52.154.131.98 (e.g. https://github.com/argotorg/solc-bin/actions/runs/29387623659).

We never passed GITHUB_TOKEN to the action, so it polls the API unauthenticated. Unauthenticated requests are limited to 60/h per IP and the runner IPs are shared across GitHub-hosted runners, so whether a run succeeds is random.

Also grants the job actions: read explicitly, since the default token permissions on this repo are restricted to contents/packages and turnstyle needs to list workflow runs.


nikola-matic:
Not sure if you still need this, but the build should be fixed 🤞

